PR #12, opened 2024-06-05
feat(data-stream): add fuel-core-nats V2

lostman:
`fuel-core-nats` is an extended version of `fuel-core`, which adds NATS publishing functionality.

See `crates/fuel-core-nats/README.md` for `fuel-core-nats` details.


.github/workflows/ci.yaml
```diff
@@ -47,6 +47,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: ${{ env.RUST_VERSION }}
+          targets: wasm32-unknown-unknown
```
pedronauck, line 50: why do we need wasm if we will not build for it?

.typos.toml
```diff
@@ -1,2 +1,2 @@
 [default]
-extend-ignore-identifiers-re = ["TRO", "tro", "Tro"]
+extend-ignore-identifiers-re = ["TRO", "tro", "Tro", "16Uiu2HAm6pmJUedRFjennk4A8yWL6zCApHCuykzRRroqMjjxZ8o6", "16Uiu2HAm8dBwTRzqazCMqQDdR8thMa7BKiW4ep2B4DoQQp6Qhyfd"]
```
pedronauck, line 2:
```suggestion
extend-ignore-identifiers-re = ["TRO", "tro", "Tro"]

[files]
extend-exclude = ["crates/**/README.md"]
```

Cargo.toml
```diff
@@ -16,6 +16,10 @@ anyhow = "1.0"
 clap = { version = "4.5", features = ["derive"] }
 futures = "0.3"
 tokio = { version = "1.38", features = ["full"] }
+fuel-core-bin = { version = "0.27", features = ["env", "p2p", "relayer", "rocksdb"] }
```
pedronauck, line 19: any specific reason you are not using the latest `fuel-core`?

crates/fuel-core-nats/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "fuel-core-nats"
```
pedronauck, line 2: can you change for `fuel-data-stream` as we talked in the last sync?

pedronauck (changes requested):

.github/workflows/ci.yaml
pedronauck: ✅

crates/fuel-core-nats/Cargo.toml
pedronauck: ok, let's go with this one as it is, then after the tech spec and more clarity about crates and etc we can bike-shedding about the name again

pedronauck (approved):

Jurshsmith (approved):

PR #19, opened 2024-06-11
build(repo): allow starting & stopping nats server locally

Jurshsmith:
This change will help set up our NATS server's local infrastructure before running `fuel-core-nats`. 

This will ease up our development setup and, more importantly, help us explicitly declare our local infrastructure dependencies, which will be useful when setting up our Kubernetes cluster.


scripts/setup.sh
```diff
@@ -1,18 +1,5 @@
...
-    if ! which "$1" >/dev/null 2>&1; then
-        echo "$1 is not installed. Please install $1 and try again."
-        exit 1
-    fi
-}
-
-# Check if Rust is installed
-check_command rustup
-# Check if pre-commit is installed
-check_command pre-commit
```
Jurshsmith, line 14: I moved the `check_command` to the top layer, the Makefile, to allow reusing.

pedronauck (approved):

scripts/setup.sh
pedronauck: ✅

PR #20, opened 2024-06-12
build(repo): automate running fuel-core-nats locally

Jurshsmith:
Introduce `start.fuel-core-nats` for starting the `fuel-core-nats` service locally


pedronauck (changes requested):

pedronauck:
@Jurshsmith, what do you think about including the local `fuel-core-nats` component in Docker Compose, configured with the correct environment variables? This way, we can provide a Docker Compose example with all the necessary tools pre-configured as I mentioned in our sync.

Jurshsmith:
@pedronauck Thanks for bringing that up again. The plan is to first prepare the Docker image for fuel-core-nats. Once that's ready, we'll create a docker-compose template like you've described. I don't think we will need this for development though, since `fuel-core-nats` is new and unstable.

Makefile
```diff
@@ -21,15 +21,15 @@ check-commands:
...
 	./scripts/setup.sh
 
 # ------------------------------------------------------------
 #  Development
 # ------------------------------------------------------------
-
-dev:
-	cargo run
+# Starts fuel-core-nats service
+start.fuel-core-nats:
```
Jurshsmith, line 31: Ended up with `start` to allow the use of `stop` DSL in the script for stopping `docker-compose` services

pedronauck (approved):

pedronauck (approved):

PR #32, opened 2024-06-24
feat(data-stream): add nkey auth

lostman:


Jurshsmith (commented):
Looks good 👍 . Could you please update the `env.sample` file with the new `NATS_NKEY` variable?

Jurshsmith:
> There's no default nkey. What should I put there?

An empty string or garbage value works

pedronauck (approved):

PR #35, opened 2024-06-27
build(data-stream): add missing default env var

pedronauck:
This PR instends to add the missing `NATS_NKEY` env var in the `.env.sample


Jurshsmith (approved):
👍🏾

PR #61, opened 2024-07-17
feat(core): add first version of streams-core crate

### LINKED ISSUE #67: Create a core crate for the project

pedronauck:
This PR introduces the `streams-core` crate, a foundational component for the entire data systems project. While it includes significant NATS functionality, its purpose extends beyond just NATS integration to provide core types, utilities, and abstractions used across various project parts.

## Key Features:
- Core type definitions and common utilities
- NATS client implementation and stream management
- Generic subject system for message routing
- Error handling and payload validation
- Reusable components for other crates in the project

## Tasks:
- [x] Implement NATS client
- [x] Add stream management functionality
- [x] Create subject system
- [x] Develop core type definitions
- [x] Write unit tests

## Next Steps:
- Integrate core functionality into other project components
- Expand documentation and usage examples
- Identify additional shared features needed across the project

⭐ Close #67


crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+        let host = container.get_host().await?;
+        let host_port = container.get_host_port_ipv4(4222).await?;
+        let url = format!("{host}:{host_port}");
+        let client = NatsClient::connect(url.as_str(), "test", None).await?;
+
+        Ok((client, async move {
+            container.stop().await?;
+            container.rm().await?;
+            Ok(())
+        }))
```
Jurshsmith, line 197: Nit:

```suggestion
        let cleanup = async move {
            container.stop().await?;
            container.rm().await?;
            Ok(())
        };
        Ok((client, cleanup))
```

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+    use super::*;
+
+    #[tokio::test]
+    async fn create_connection() -> BoxedResult<()> {
+        let (client, cleanup) =
+            NatsClient::connect_with_testcontainer().await?;
+        assert!(client.is_connected().await?);
+        cleanup.await?;
+        Ok(())
+    }
```
Jurshsmith, line 212: Fwiw, this doesn't test our actual connection.

Jurshsmith: Also, I'm not sure why we need TestContainers when we already contextualized the connections with subject prefixes.

crates/core/src/nats/streams.rs
```diff
@@ -0,0 +1,189 @@
...
+impl StreamKind {
+    pub fn get_subjects(&self, prefix: &str) -> Vec<String> {
+        match self {
+            Self::Blocks => vec![SubjectName::Blocks.with_prefix(prefix)],
+            Self::Transactions => vec![
+                SubjectName::Transactions.with_prefix(prefix),
+                SubjectName::TransactionsById.with_prefix(prefix),
+            ],
+        }
+    }
```
Jurshsmith, line 35: Why do we need this? It's a leak with our Enum abstraction. If this is how we have to manage the subjects, then we should probably revisit the entire Subjects abstraction.

crates/core/src/nats/streams.rs
```diff
@@ -0,0 +1,189 @@
...
+    },
+    NatsError,
+    SubjectName,
+};
+
+#[derive(Debug, EnumIter, Clone, Hash, Eq, PartialEq)]
+pub enum StreamKind {
+    Blocks,
+    Transactions,
+}
```
Jurshsmith, line 24: I am not sure why we decided to separate `StreamKind` from `SubjectName`. The complexity grows significantly doing this.

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+    ) -> Result<async_nats::Client, NatsError> {
+        let options = async_nats::ConnectOptions::new()
+            .connection_timeout(Duration::from_secs(30))
+            .max_reconnects(10);
+
+        if let Some(nkey) = nkey {
+            async_nats::connect_with_options(&url, options.nkey(nkey)).await
+        } else {
+            async_nats::connect_with_options(&url, options).await
+        }
```
Jurshsmith, line 78: Do we want the possibility of connecting without `nkey` just for tests? I think we should rethink using TestContainers.

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    pub async fn create_conn(
+        url: &str,
+        nkey: Option<String>,
+    ) -> Result<async_nats::Client, NatsError> {
+        let options = async_nats::ConnectOptions::new()
+            .connection_timeout(Duration::from_secs(30))
+            .max_reconnects(10);
+
```
Jurshsmith, line 73: Nice 👍

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+        match conn_state {
+            connection::State::Pending => {
+                Err(NatsError::ConnectionPending(self.url.to_owned()))
+            }
+            connection::State::Disconnected => {
+                Err(NatsError::ConnectionDisconnected(self.url.to_owned()))
+            }
+            connection::State::Connected => Ok(true),
+        }
+    }
```
Jurshsmith, line 96: With the connection strategy you've written above, and our pod restart policy, we can do away with this to reduce complexity.

crates/core/src/nats/streams.rs
```diff
@@ -0,0 +1,189 @@
...
+            ],
+        }
+    }
+
+    pub fn get_storage(&self) -> NatsStorageType {
+        match self {
+            Self::Blocks => NatsStorageType::File,
+            Self::Transactions => NatsStorageType::File,
+        }
+    }
```
Jurshsmith, line 42: I don't see how all storage won't be `NatsStorageType::File`? To keep it simple, we can avoid this function altogether

crates/core/src/nats/streams.rs
```diff
@@ -0,0 +1,189 @@
...
+        match self {
+            Self::Blocks => "blocks",
+            Self::Transactions => "transactions",
+        }
+    }
+
+    pub fn get_stream_config(&self, prefix: &str) -> JetStreamConfig {
+        JetStreamConfig {
+            subjects: self.get_subjects(prefix),
+            storage: self.get_storage(),
```
Jurshsmith, line 54:
```suggestion
            storage: NatsStorageType::File,
```

crates/core/src/nats/streams.rs
Jurshsmith: A better abstraction might be to get merge of `Subjects` & `StreamKind` by creating an Enum type for each Stream:

`pub enum BlocksStream { Blocks }`

`pub enum TransactionsStream { Transactions, TransactionsById {...} }`

What do you think?

crates/core/src/nats/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Nats {
+    pub client: NatsClient,
+    pub streams: Streams,
+}
+
+impl Nats {
+    pub async fn new(
+        conn_id: String,
```
Jurshsmith, line 21: Nit:

```suggestion
        conn_id: &str,
```

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+        if payload_size > max_payload_size {
+            return Err(NatsError::PayloadTooLarge {
+                subject: subject_name.to_string(),
+                payload_size,
+                max_payload_size,
+            });
+        }
+
+        Ok(())
+    }
```
Jurshsmith, line 64: We should avoid defensively validating things that are outside our system. The NATS cluster is a separate system and should own the logic of validating the payload size since anyone can try that out.

In short, this doesn't defend anything but adds to the runtime cost. We should instead handle this error on publish and report it appropriately.

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,262 @@
...
+impl NatsClient {
+    pub async fn connect(
+        url: &str,
+        conn_id: &str,
+        nkey: Option<String>,
+    ) -> Result<Self, NatsError> {
+        let conn_id = conn_id.to_string();
+        let conn = Arc::new(Self::create_conn(url, nkey).await?);
+        let context =
+            Arc::new(async_nats::jetstream::new(conn.as_ref().clone()));
```
Jurshsmith, line 37: Why do we want the `NatsClient` to be clonable?

Jurshsmith (changes requested):

crates/core/src/nats/streams.rs
pedronauck: My idea was that maybe in the future we could have some request-reply, but of course, we can go in a simpler approach

crates/core/src/nats/client.rs
pedronauck: There's a bug related to the `jetstream::new(client)` method that just worked when using like this, I tried many different approaches and in the end this was the one that worked 😕

crates/core/src/nats/client.rs
pedronauck: Ok, I removed it from the `publish` inside the `NatsClient` and just left the method here so you can use it later in the Publisher within `fuel-core-nats`. Is this what you meant to do?

crates/core/src/nats/client.rs
pedronauck: ✅

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,223 @@
...
+            url: url.to_string(),
+            conn_id,
+            conn,
+            jetstream: context,
+        })
+    }
+
+    async fn create_conn(
+        url: &str,
+        nkey: Option<String>,
```
Jurshsmith, line 49: Why not:

```suggestion
        nkey: &str,
```

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        let conn =
+            async_nats::connect_with_options(&url, options.nkey(nkey.unwrap()))
+                .await
+                .map_err(|e| NatsError::ConnectError {
+                    url: url.to_owned(),
+                    source: e,
+                })?;
+
+        Ok(Arc::new(conn))
```
Jurshsmith, line 69: Do we need the `Arc`?

```suggestion
        Ok(conn)
```

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    pub async fn publish(
+        &self,
+        subject: Subject,
+        payload: Bytes,
+    ) -> BoxedResult<&Self> {
+        let subject_prefixed = subject.with_prefix(&self.conn_id);
+        let context = self.jetstream.as_ref();
+        let ack_future = context.publish(subject_prefixed, payload).await?;
```
Jurshsmith, line 97: Sorry, what I meant was that we can get rid of `validate_payload`, and then match on this future's error to return log an error if the payload is too large.

crates/core/src/nats/client.rs
```diff
@@ -0,0 +1,223 @@
...
+        if payload_size > max_payload_size {
+            return Err(NatsError::PayloadTooLarge {
+                subject: subject_name.to_string(),
+                payload_size,
+                max_payload_size,
+            });
+        }
+
+        Ok(self)
+    }
```
Jurshsmith, line 88:
```suggestion

```

crates/core/src/nats/streams.rs
```diff
@@ -0,0 +1,176 @@
+use std::collections::HashMap;
```
Jurshsmith, line 1: As discussed, we need a middle ground for the Streams and Subjects. That'll be in my next PR.

Jurshsmith (commented):
Looks good 👍🏾. Just a few comments I should probably take on for my next PR.

crates/core/src/nats/client.rs
pedronauck: If you look below, it's returning an error when nkey is none

crates/core/src/nats/client.rs
pedronauck: I mentioning in the previews comment related to this why

crates/core/src/nats/client.rs
pedronauck: There's no payload size error related to the publish futures, I don't see this as a huge resource-consuming or complex algorithm, it's not introducing high level of complexity or something like that, this could be improved to have `max_payload_size` in the struct to avoid get this from the server each time, but anyway, isn't something to go further, I don't see problems to stay with it here

CleanShot 2024-07-24 at 17 38 10

crates/core/src/nats/client.rs
pedronauck: I will solve this conversation since we are already talking about this in the other one

crates/core/src/nats/streams.rs
pedronauck: ✅

crates/core/src/nats/client.rs
Jurshsmith: Yes, that's the point. We don't need to worry about that error if we type-enforce that an NKEY must always be provided

crates/core/src/nats/client.rs
Jurshsmith: `Arc` is usually used as a wrapper type. Returning it from a factory function doesn't seem right. The caller should get to decide if they want to wrap it with an `Arc`, `Rc`, or `Box`

crates/core/src/nats/client.rs
pedronauck: **tldr;** 
I tested here, and it's impossible to do what you propose. In fact, it's better to always check the payload using the `publish` method since it's just a simple conditional check.

First, as you can see, there's no error related to payload size in the publish future's error enum:
CleanShot 2024-07-24 at 17 56 13

Second, if you try to decrease the `max_payload` in the `nats.conf`, you will see that by doing that and publishing anything, there's no "max payload error"; the error that shows on Nats is a timeout error as below.

```
**Error: CreateStreamFailed { name: "connection-1437309941_stream:blocks", source: Error { kind: TimedOut, source: None } }**
```

Apart from these points, as mentioned earlier, I don't see a major issue here because this method is quite simple. There's no asynchronous processing involved (please refer to the most updated version), and the amount of processing is minimal, with almost no added complexity. I don't foresee this becoming a performance issue in the future. However, we did encounter difficulties with this validation when Maciej was testing, which is why he introduced this method. If we have concerns, we can review this and explore potential solutions. For now, in my humble opinion, we can proceed as planned.

Jurshsmith: > Second, if you try to decrease the max_payload in the nats.conf, you will see that by doing that and publishing anything, there's no "max payload error"; the error that shows on Nats is a timeout error as below.

Yes, that still works. We can match on the `TimedOut` error and then check if the payload was greater than our max payload to deduce what gets logged. 

>  I don't foresee this becoming a performance issue in the future. However, we did encounter difficulties with this validation when Maciej was testing, which is why he introduced this method. If we have concerns, we c

This is more of a readability concern than a performance optimization. The success path for publishing should be very minimal when read making it easy to point out exceptions.

> we can review this and explore potential solutions. For now, in my humble opinion, we can proceed as planned.

Yes, I have no worries merging this. Just wanted to emphasize the preferred paradigm of avoiding defensive programming until when necessary (i.e. for unstructured inputs).

crates/core/src/nats/client.rs
pedronauck: Got your point, makes sense

crates/core/src/nats/client.rs
pedronauck: I tried again without `Arc` and this time it worked, maybe the bug was something not related to the pointer itself when I tried earlier

Jurshsmith (approved):

PR #103, opened 2024-07-29
refactor(repo): add AsRef trait for api arguments

0xterminator:
This small PR adds:

- a more ergonomic interface for input args which can now implement AsRef<str> allowing them to be of types String, &str, Cow<str> etc as well as any other AsRef impl

- removes one unwrap and maps it to a proper error message

### DIFF
--- a/crates/core/src/nats/streams.rs
+++ b/crates/core/src/nats/streams.rs
@@ -24,7 +24,8 @@ pub enum StreamKind {
 }
 
 impl StreamKind {
-    pub fn get_subjects(&self, prefix: &str) -> Vec<String> {
+    pub fn get_subjects(&self, prefix: impl AsRef<str>) -> Vec<String> {
+        let prefix = prefix.as_ref();
         match self {
             Self::Blocks => vec![SubjectName::Blocks.with_prefix(prefix)],
             Self::Transactions => vec![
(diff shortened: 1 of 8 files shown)


crates/core/src/nats/streams.rs
```diff
@@ -24,7 +24,8 @@ pub enum StreamKind {
 }
 
 impl StreamKind {
-    pub fn get_subjects(&self, prefix: &str) -> Vec<String> {
+    pub fn get_subjects(&self, prefix: impl AsRef<str>) -> Vec<String> {
```
pedronauck, line 27: Since this struct will be completely removed in favor of the new implementation, can you remove the changes here to avoid unnecessary conflicts?

crates/core/src/nats/subjects.rs
```diff
@@ -11,13 +11,13 @@ pub enum SubjectName {
 }
 
 impl SubjectName {
-    pub fn with_prefix(&self, prefix: &str) -> String {
-        [prefix, ".", &self.to_string()].concat()
+    pub fn with_prefix(&self, prefix: impl AsRef<str>) -> String {
```
pedronauck, line 14: You can remove the changes here also, this will also be removed

pedronauck (changes requested):

Jurshsmith (changes requested):
Nice initiative for improving the interfaces with the AsRef trait. 

However, I think it might actually be more ergonomic to stick with `&str`. I would only recommend using the AsRef trait when shipping public interfaces, that way, end users can pass whatever argument they want -- this will be very applicable in `fuel-streams-sdk`. 

The following highlights why I think `&str` is mostly better:

- `&str` will consistently compile faster than `AsRef`
-  AsRef requires more transformations before being consumed. Since it's an application code,  it would require more repeated code to use `AsRef`'s everywhere. (notably the extra `.as_ref()` calls)

What do you think?

0xterminator:
Closing because of current changes taking place.

Issue #95, opened 2024-07-24
Setup cargo audit


pedronauck:
## Description
Integrate cargo audit into the project workflow to automatically check for security vulnerabilities in dependencies.

## Tasks
- [x] Install cargo audit in the development environment
- [x] Add cargo audit to the CI pipeline
- [x] Configure cargo audit settings (if necessary)
- [x] Create a process for regularly running cargo audits and addressing findings (get inspiration from what [we have in fuel-core](https://github.com/fuellabs/fuel-core))

## Technical Details
Use cargo audit from https://github.com/rustsec/cargo-audit

## Additional Notes
Consider setting up scheduled runs of cargo audit to catch new vulnerabilities promptly.

## Definition of Done
- [ ] Code implemented
- [ ] Unit tests written and passing
- [ ] Integration tests (if applicable) written and passing
- [ ] Documentation updated
- [ ] Code reviewed

PR #104, opened 2024-07-29
build(repo): added cargo-audit to CI/CD and project

0xterminator:
This MR implements the scope defined under: https://github.com/FuelLabs/data-systems/issues/95


pedronauck:
My only question is: should we run this at the pull request triggering level or just after merges on master?

.github/workflows/ci.yaml
```diff
@@ -113,6 +113,15 @@ jobs:
       - name: Running linting
         run: make lint
 
+  audit:
```
pedronauck, line 116: My only question is: should we run this at the pull request triggering level or just after merges on master?

0xterminator: I would run it for now a the PR level as w have not that many deps. Maybe later we can do it on master merge ? Let me know what you think.

pedronauck: It's better to add it to the right place to avoid creating tech debt for the future. We can follow also what `fuel-core` has: https://github.com/FuelLabs/fuel-core/blob/master/.github/workflows/nightly-cargo-audit.yml and it as a separate Github action.

0xterminator: @pedronauck  I have added the nightly-cargo-audit similarly to fuel-core. Now I have removed the audit step from the ci.yaml

pedronauck (approved):

pedronauck (approved):

PR #105, opened 2024-07-29
fix(publisher): publish compact json rather than pretty json

Jurshsmith:
Currently, we are publishing our subjects and payloads using pretty json. This quickly added up and caused a large ROM consumption in our NATS cluster. For obvious reasons, pretty json is way heavier compared with it's compact equivalent. This change fixes this by only publishing the blocks and transactions serialized to compact json.

### DIFF
--- a/crates/fuel-core-nats/src/lib.rs
+++ b/crates/fuel-core-nats/src/lib.rs
@@ -158,7 +158,7 @@ impl Publisher {
 
         // Publish the block.
         info!("NATS Publisher: Block#{height}");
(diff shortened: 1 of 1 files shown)


0xterminator (approved):
should def be a bit of improvement!

pedronauck:
Can you just fix the checks first? I guess publisher is not configured as scope already in the workflow

Jurshsmith:
@pedronauck all checks have passed.

pedronauck (approved):

Issue #28, opened 2024-06-18
Set up NATS message deduplication


lostman:
### Motivation

We can use NATS's built-in deduplication feature to ensure each message is delivered only once.

For that, we need to set the message ID.

Since each message we publish is uniquely identified by it's full topic, I think we can simply set the message_id to be the same as topic (.e.g. `blocks.12345`).

In Rust, the call to publish (with a custom message ID) will look like:

```
    /// let ack = jetstream
    ///     .send_publish(
    ///         "events",
    ///         Publish::build().payload("data".into()).message_id("uuid"),
    ///     )
    ///     .await?;
```

See

https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive#message-deduplication

for more details.

pedronauck:
Closed on #107

PR #107, opened 2024-07-29
feat(core): added nats publishing deduplication logic and tests

0xterminator:
Issue: https://github.com/FuelLabs/data-systems/issues/28

### DIFF
--- a/crates/fuel-core-nats/src/nats.rs
+++ b/crates/fuel-core-nats/src/nats.rs
@@ -1,6 +1,6 @@
 mod subjects;
 use anyhow::Context;
-use async_nats::jetstream::stream;
+use async_nats::jetstream::{context::Publish, stream};
 pub use subjects::*;
 use tracing::info;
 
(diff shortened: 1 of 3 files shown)


crates/fuel-core-nats/src/nats.rs
```diff
@@ -37,7 +37,12 @@ impl NatsConnection {
             )
         }
         // Publish
-        let ack_future = self.jetstream.publish(subject, payload).await?;
+        let mut headers_map = HeaderMap::new();
+        headers_map.append(NATS_MESSAGE_ID, subject.to_string().as_str());
```
pedronauck, line 41: what is this `HeaderMap` for?

0xterminator: The headermap sets some dedup headers which is btw. achieved with .message_id(...) which I have now switched to here. Sorry, I did not see the method has a message_id.

pedronauck (approved):

PR #108, opened 2024-07-29
refactor(core): stream and subjects implementation

### LINKED ISSUE #101: Improve subjects and streams architecture on SDK
## Description

Currently, `SubjectName`, `Subject`, and `StreamKind` within the core crate create unnecessary complexity and confusion; the idea is to create a trait for streams and, for each stream, a related enum.

## Technical Details
- Each stream will have their own sets of subjects
- The stream and the subjects need to be able to add a prefix on their names

## Definition of Done
- [ ] Code implemented
- [ ] Unit tests written and passing
- [ ] Integration tests (if applicable) written and passing
- [ ] Documentation updated
- [ ] Code reviewed

pedronauck:
⭐ Closes #101


0xterminator:
@pedronauck You might want to rebase against main as there are conflicts.

crates/core/src/nats/streams/stream.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use strum::IntoEnumIterator;
+
+use crate::nats::{types, NatsClient, NatsError};
+
+pub trait StreamIdentifier {
+    const STREAM: &'static str;
+}
+
+pub trait StreamSubjectsEnum:
```
0xterminator, line 11: I feel the enum in the title makes it confusing. Why not just StreamSubjects ?

0xterminator (dismissed):

pedronauck:
> @pedronauck You might want to rebase against main as there are conflicts.

I already did, btw, I just moved the dedup test you created to `tests/` package

crates/core/src/nats/conn_streams.rs
```diff
@@ -0,0 +1,70 @@
...
+    subjects::{BlockSubjects, TransactionSubjects},
+    NatsClient,
+    NatsError,
+};
+
+#[derive(Debug, Clone)]
+pub struct ConnStreams {
+    pub blocks: Stream<BlockSubjects>,
+    pub transactions: Stream<TransactionSubjects>,
+}
```
Jurshsmith, line 12: Since Blocks & Transactions are variants of a stream, it would be better to express them as Enums

```suggestion
pub enum ConnStream {
    Blocks(Stream<BlockSubjects>),
    Transactions(Stream<TransactionSubjects>),
}
```

crates/core/src/nats/conn_streams.rs
```diff
@@ -0,0 +1,70 @@
...
+            blocks,
+        })
+    }
+}
+
+#[cfg(any(test, feature = "test_helpers"))]
+impl ConnStreams {
+    pub fn get_stream_list(&self) -> Vec<super::types::AsyncNatsStream> {
+        vec![self.blocks.stream.clone(), self.transactions.stream.clone()]
+    }
```
Jurshsmith, line 30: This would just be an iteration of the streams Enum

crates/core/src/nats/conn_streams.rs
```diff
@@ -0,0 +1,70 @@
...
+        streams: Vec<super::types::AsyncNatsStream>,
+    ) -> anyhow::Result<Vec<String>> {
+        let mut all_subjects = Vec::new();
+        for mut stream in streams {
+            let info = stream.info().await?;
+            all_subjects.extend(info.config.subjects.clone());
+        }
+
+        Ok(all_subjects)
+    }
```
Jurshsmith, line 42: Nit: Not sure why this is embedded in `ConnStream`'s impl (?) Since it's a test helper that takes an AsynNatsStream type

crates/core/src/nats/nats_conn.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+impl NatsConn {
+    pub async fn new(
+        conn_id: &str,
+        nats_url: &str,
+        nats_nkey: &str,
+    ) -> Result<Self, NatsError> {
+        let client = NatsClient::connect(nats_url, conn_id, nats_nkey).await?;
+        let streams = ConnStreams::new(&client).await?;
```
Jurshsmith, line 16: We just need a list of streams, not another struct. What do you think?

```suggestion
        let streams = streams:get_all(&client);
```

Jurshsmith (dismissed):
Great work 🚀. It looks so much better. 

I have a tiny interface change for the `Streams` collection. However, it's not a blocker -- it looks good overall. Thanks for taking the time to rework this!

crates/core/src/nats/client.rs
```diff
@@ -31,8 +31,8 @@ impl NatsClient {
         nkey: &str,
     ) -> Result<Self, NatsError> {
         let conn_id = conn_id.to_string();
-        let conn = Self::create_conn(url, nkey).await?;
-        let context = async_nats::jetstream::new(conn.clone());
+        let conn = Self::create_conn(url, nkey, conn_id.as_str()).await?;
```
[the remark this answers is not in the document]
pedronauck: Oh that's new, nice 👏🏻

crates/core/src/nats/conn_streams.rs
pedronauck: This isn't possible because I need to keep the streams in some place in the memory since they are created at runtime, so if I added this here as a enum, I'll need another struct to save the streams, so kinda doesn't make sense

crates/core/src/nats/conn_streams.rs
pedronauck: It's been used outside of this file within the `TestStreamBuilder`, but I'll change it indeed to simplify its usage

Jurshsmith (approved):

Jurshsmith (dismissed):

Jurshsmith (dismissed):

Jurshsmith (approved):

PR #121, opened 2024-08-05
refactor(core): model `Streamable` trait for `fuel-core` types directly for simplicity

Jurshsmith:
This PR introduces the `Streamable` trait for `fuel-core` types, i.e. `Block`, Transactions, etc. to simplify the DX when 
publishing and building stream queries. This also shows the explicit correlation between our APIs and the `fuel-core-types`.

With this change, we can now easily think of the core types (Blocks/Transactions) from `fuel-core-type` as streams with their Subjects and Streamable behaviors defined in our `fuel-streams-core` library.


pedronauck (approved):

pedronauck:
Nice job @Jurshsmith 👏🏻

PR #123, opened 2024-08-08
refactor(core): use kv store instead of stream

pedronauck:
## Description

- This PR changes our data store mechanism from streams to use KV store.
- Also, this PR introduces a new derive macro for subjects, adding some methods on them:
    - `Subject::all()` - show the wildcard to query all 
    - `Subject::wildcard(..params)` - autogenerate method for build a wildcard query with struct params
    - `Subject::with_<fields>` - autogenerate methods do set struct fields
    - `#[derive(Subject)]` - the derive macro to use in the struct

## Changes
- [x] change from `ConnStreams -> ConnStores`
- [x] change from `Stream<S: Streamable> -> Store<S: Storable>`
- [x] `client.publish()` removed because now has the `store.upsert()`
- [x] change from `client.create_stream() -> client.create_store()`
- [x] `client.create_consumer` removed because now has the `store.watch()`


crates/fuel-streams-core/src/nats/conn_stores.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::{
+    nats::{NatsClient, NatsError},
+    types::{Block, Transaction},
+};
+
+#[derive(Clone)]
+pub struct ConnStores {
+    pub blocks: Store<Block>,
+    pub transactions: Store<Transaction>,
+}
```
Jurshsmith, line 11: This public API from `fuel-streams-core` will allow streaming blocks and transactions. why call it it `Stores`?

crates/fuel-streams-core/src/nats/nats_client.rs
```diff
@@ -64,46 +51,39 @@ impl NatsClient {
         Ok(self)
     }
 
-    pub async fn create_stream(
+    pub async fn create_store(
```
Jurshsmith, line 54: This makes sense since it's a Nats concern. to even make it more specific, we can rename as: `create_kv_store`

crates/fuel-streams-core/src/nats/store.rs
```diff
@@ -0,0 +1,90 @@
...
+    kv::Watch,
+};
+use async_trait::async_trait;
+use fuel_streams_macros::subject::IntoSubject;
+
+use super::{NatsClient, NatsNamespace, StoreError};
+use crate::nats::{types::*, NatsError};
+
+#[async_trait]
+pub trait Storable: Debug + Clone + serde::Serialize {
```
Jurshsmith, line 14: `Streamability` was a concern of `fuel-streams-core`, not `nats`. That's why I mentioned taking these traits, and subjects to the root level. So, renaming it to `Storable` just because we are using the kv store internally is a leak.

crates/fuel-streams-core/src/nats/store.rs
```diff
@@ -0,0 +1,90 @@
...
+    const STORE: &'static str;
+
+    /// This is temporary until we have the data parser
+    fn store_encode(&self) -> Result<Vec<u8>, StoreError> {
+        bincode::serialize(self).map_err(StoreError::SerializationFailed)
+    }
+
+    /// This is temporary until we have the data parser
+    fn store_decode<'a>(val: &'a [u8]) -> Result<Self, StoreError>
```
Jurshsmith, line 23: prefixing these with `store_` is even more confusing since encoding and decoding is totally agnostic of the `store`

crates/fuel-streams-core/src/nats/store.rs
```diff
@@ -0,0 +1,90 @@
...
+    where
+        Self: serde::Deserialize<'a>,
+    {
+        bincode::deserialize(val).map_err(StoreError::SerializationFailed)
+    }
+
+    async fn create_store(
```
Jurshsmith, line 30: This is fine (`create_kv_store`), since for a `fuel-core-type` to be a `Streamable` we need a nice helper function to create a KV store, but we could go further by taking it away from this public trait and making private.

Issue #97, opened 2024-07-24
Refactor to use core crate in the publisher crate


pedronauck:
## Description
Refactor the publisher crate to utilize the new core::nats module, improving code organization and reducing duplication.

## Tasks
- [ ] Update publisher crate dependencies to include the core crate
- [ ] Replace existing NATS-related code in publisher with core::nats module
- [ ] Update affected tests in the publisher crate

## Technical Details
Use the new core::nats module for NATS-related functionality in the publisher crate.

## Additional Notes
This refactoring will improve maintainability and ensure consistent NATS handling across the project.

## Definition of Done
- [ ] Code implemented
- [ ] Unit tests written and passing
- [ ] Integration tests (if applicable) written and passing
- [ ] Documentation updated
- [ ] Code reviewed

PR #124, opened 2024-08-13
feat(core): refactor core and publisher to integrate coherently

Jurshsmith:
This PR introduces key architectural enhancements and refactors the core to better integrate with fuel-streams-publisher. The changes focus on improving modularity, clarity, and flexibility across the codebase:

- **Streamable Trait Generalization**: The Streamable trait is now NATS-agnostic by adopting more generic function signatures. This decouples our core logic from specific messaging systems, allowing for easier future integrations with other stream processing technologies without significant refactoring.

- **Improved Documentation and Abstraction**: Existing abstractions are now more clearly documented, promoting well-defined separation of concerns. This makes it easier for contributors to understand and extend the architecture while reducing the likelihood of cross-cutting concerns.

- **Decoupled Stream Types**: This PR introduces the Streams crate in both fuel-streams-publisher and fuel-streams. By separating streams that can be published from those that can be consumed, the architecture is more flexible and better aligned with the principle of single responsibility, leading to cleaner, more maintainable code.

Closes: https://github.com/FuelLabs/data-systems/issues/83
Closes: https://github.com/FuelLabs/data-systems/issues/97
Closes: https://github.com/FuelLabs/data-systems/issues/84


crates/fuel-streams-core/src/blocks/subjects.rs
```diff
@@ -10,16 +10,13 @@ pub struct BlocksSubject {
...
-    }
-}
-
 impl From<&Block> for BlocksSubject {
     fn from(block: &Block) -> Self {
         let block_height = *block.header().height();
-        BlocksSubject::new().with_height(Some(BlockHeight::from(block_height)))
+        BlocksSubject::new()
+        // TODO: Use correct block producer here
+            .with_producer(Some("0x".to_string()))
```
0xterminator, line 18: Question: I never got where the block producer is supposed to come - is this something we can extract form the Block or is it external data that needs to passed in here ?

crates/fuel-streams-core/src/nats/nats_client.rs
```diff
@@ -6,60 +6,44 @@ use async_nats::{
...
             nats_client,
             jetstream,
             opts: opts.to_owned(),
-            namespace: opts.namespace,
+            namespace,
         })
     }
 
-    #[allow(dead_code)]
-    fn validate_payload(
```
0xterminator, line 43: Why are we dropping this ?

crates/fuel-streams-core/src/nats/nats_client_opts.rs
```diff
@@ -46,23 +44,18 @@ impl ClientOpts {
         Self { role, ..self }
     }
 
-    #[cfg(any(test, feature = "test-helpers"))]
```
0xterminator, line 49: Is the conn Id no longer needed and overtaken by the namespace ?

crates/fuel-streams-core/src/stream.rs
```diff
@@ -0,0 +1,230 @@
...
+    + Sync
+    + DataParserSerializable
+    + DataParserDeserializable
+    + 'static
+{
+    async fn encode(&self, subject: &str) -> Vec<u8> {
+        Self::data_parser()
+            .to_nats_payload(subject, self)
+            .await
+            .expect("Streamable must encode correctly")
```
0xterminator, line 41: I feel we should handle the error here rather than unwrap it ?

crates/fuel-streams-core/src/stream.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    async fn decode(encoded: Vec<u8>) -> Self {
+        Self::decode_raw(encoded).await.data
+    }
+
+    async fn decode_raw(encoded: Vec<u8>) -> NatsFormattedMessage<Self> {
+        Self::data_parser()
+            .from_nats_message(encoded)
+            .await
+            .expect("Streamable must decode correctly")
```
0xterminator, line 52: ditto here

crates/fuel-streams-core/src/transactions/mod.rs
```diff
@@ -8,13 +8,11 @@ use types::*;
 use crate::prelude::*;
 
 impl StreamEncoder for Transaction {}
+
 impl Streamable for Transaction {
     const NAME: &'static str = "transactions";
     const WILDCARD_LIST: &'static [&'static str] = &[
```
0xterminator, line 14: Can we not introduce back the .all() here that would return the list of all Subjects::Wildcards ? ... rather than enumerating them

crates/fuel-streams-core/src/transactions/types.rs
```diff
@@ -55,12 +55,14 @@ impl From<Transaction> for TransactionKind {
...
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub enum TransactionStatus {
     Failed,
     Submitted,
     SqueezedOut,
     Success,
+    #[default]
+    None,
```
0xterminator, line 65: Is there a none status defined in fuel-core ? I thought its also either failed, submitted, squeezed out or successful ?

crates/fuel-streams-publisher/src/transactions.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    for (transaction_index, transaction) in transactions.iter().enumerate() {
+        let tx_id = transaction.id(chain_id);
+        let kind = TransactionKind::from(transaction.to_owned());
+        let status: TransactionStatus = off_chain_database
+            .get_tx_status(&tx_id)?
+            .map(|status| status.into())
+            .unwrap_or_default();
+
+        let transactions_subject: TransactionsSubject =
```
0xterminator, line 35: Cant we do From trait here to just use .into() ?

crates/fuel-streams-core/src/blocks/subjects.rs
Jurshsmith: I haven't investigated deeply on this yet but I intend deferring for a follow-up PR to avoid delaying this one any longer. Of course, I am open to any ideas. 🙂

crates/fuel-streams-core/src/nats/nats_client.rs
Jurshsmith: It's not used anywhere atm and it would be a concern of `fuel-streams-publisher`, so it should not exist in the `core`

crates/fuel-streams-core/src/nats/nats_client_opts.rs
Jurshsmith: Yes, correct 💯

crates/fuel-streams-core/src/stream.rs
Jurshsmith: Since we use explicit error handling, I am trying to reduce the complexity of our systems by making anomalies obvious with unwraps and useful errors should remain handled explicitly. What do you think?

crates/fuel-streams-core/src/transactions/mod.rs
Jurshsmith: Any reason why? The `.all()` just seems like an extra indirection

crates/fuel-streams-core/src/transactions/types.rs
Jurshsmith: Yeah, it's an Option<TransactionStatus> type. Unfortunately, our `IntoSubject` macro doesn't currently support Option<T> types. This might be a reason to want to revert to a simple trait (?). 

This is a normalized solution and it sort of makes sense since we want to ingest the status string, nothing else. What do you think?

crates/fuel-streams-publisher/src/transactions.rs
Jurshsmith: No, it won't work that way since the source type is an `Option<T>`

crates/fuel-streams-core/src/blocks/subjects.rs
0xterminator: Thats fine, was just a question as I am not clear on it either :) Lets keep the conversation so we can revisit it later in another MR.

crates/fuel-streams-core/src/nats/nats_client.rs
0xterminator: Right, does it make sense to move it to the publisher then ? Just asking as the check itself makes total sense and we might need it at some point ?

crates/fuel-streams-core/src/nats/nats_client_opts.rs
0xterminator: got it! All good.

crates/fuel-streams-core/src/stream.rs
0xterminator: to nats_payload throws errors that come about compaction/serialization etc. I believe you could at least cathc it and print it here with error! so we can see it before panicking. I believe the error should be handled here, but do not insist on that tbh. Your call!

crates/fuel-streams-core/src/transactions/mod.rs
0xterminator: Just asking, saw @pedronauck had it and you removed. I dont mind the current approach.

crates/fuel-streams-core/src/transactions/types.rs
0xterminator: Oh, yes, sorry, Did not see its an option type. Then you need the default yes.

crates/fuel-streams-publisher/src/transactions.rs
0xterminator: Not sure what you mean, isnt this just a mapping of T -> S. Event with options you couldo do: .map(Into::into)

Jurshsmith: Oh, sorry. Missed the line you were referring to. My earlier comment was about `TransactionStatus`. 

For this subject, we need external data to compose it correctly, namely, `chain_id`, `block_height`, et al, and they can't be gotten from `Transaction`. That said, we do need a better abstraction for these transformations, but it is one of the things I decided to defer. 👍🏾

crates/fuel-streams-core/src/nats/nats_client.rs
Jurshsmith: Not at the moment, it's currently more ergonomic to delegate that logic to the NATS infrastructure. We control both the publishing and the Stream infrastructure, so there's no reason we need to validate at runtime before publishing since that's a strong indication of a bad configuration from our side (One of the anomalies I mentioned earlier that we can avoid)

crates/fuel-streams-core/src/stream.rs
Jurshsmith: The idea is that we can't recover if encoding and decoding fail. Since we control the source data, and how it's been encoded, it is our responsibility to ensure that these must always succeed. If it doesn't, our system goes into an irrecoverable state and it should simply panic.

Please see: https://doc.rust-lang.org/beta/book/ch09-03-to-panic-or-not-to-panic.html#:~:text=The%20panic!,your%20code%20could%20recover%20from.

crates/fuel-streams-publisher/src/transactions.rs
0xterminator: OK, fair enough with me!

crates/fuel-streams-core/src/stream.rs
0xterminator: Right, but I still wonder: are we not allowed to error on malformed payload and then simply continue ? One messages might be lost but we wont stop the entire publishing ? Its more of a design decision maybe ?

0xterminator (approved):

crates/fuel-streams-core/src/stream.rs
Jurshsmith: Yes, exactly. We are publishing sensitive data possibly used in DeFi applications, so my design approach so far has been in favor of 100% correctness first.

Jurshsmith: > One messages might be lost but we wont stop the entire publishing ?

We should stop publishing since the stream chains possibly dependent data. We can only decouple concerns like that by `chain_id` since streams from Ethereum are different from Solana for example.

PR #143, opened 2024-08-15
refactor(core): use streamer instead of stream

pedronauck:
# Description

I'm just rolling back what I did before [this PR being merged](https://github.com/FuelLabs/data-systems/pull/124) and bringing this as a discussion to put it back in the code.

### There are only two things on this PR:

- [x] Adjust the `stream.rs` file according to other modules in the core crate:
It's the only one that has a different shape from other modules.

- [x] Rename the exposed `Stream` crate to `Streamer`
I renamed it before because it can easily create confusion with another stream struct we have across the project (futures `Stream` and SDK `Stream), and to make it work and get rid of this confusion, some extra workarounds, as @AJ did need to be implemented to don't have conflicts. So, to keep things standard and avoid extra work, makes sense to rename it to `Streamer`


Jurshsmith (commented):
Looks good. 

On the renaming of `Stream` to `Streamer`, I would suggest renaming the public `Stream` API exposed from `fuel-streams` as `PublicStream`, and then, re-export it with the alias `Stream`. 

This might be better because:

- Explicitly naming it `PublicStream` makes us cautious of the APIs we declare there
- We will read the `Stream` internally more and this will make maintenance a lot easier if the common `Stream` from the `core` is what depicts our Streams.
- The internal `Stream` and `PublicStream` are both streams that might implement the standard futures `Stream`

Please let me know what you think 🙏🏾

pedronauck:
Makes sense @Jurshsmith, I'll do this when I get my computer - or if you wanna do the change in the meantime, it can be also. If you change, just warn then so I can change the PR for ready for review.

Jurshsmith:
As discussed, I ended up keeping the Stream name in `fuel-streams` to better serve our end users when they inspect the SDK's APIs. For the namespace's ergonomics, I've re-exported the core's Stream interfaces directly from `fuel_streams_core` so that we can access it via the namespace directly as we do for external crates (e.g `futures_util::Stream`)

PR #144, opened 2024-08-16
ci(repo): fix release and pre-release process

pedronauck:
Closes [DS-36](https://linear.app/fuel-network/issue/DS-36/fix-release-process-with-knope)


crates/fuel-streams/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "fuel-streams"
-authors = { workspace = true }
-keywords = { workspace = true }
-edition = { workspace = true }
-homepage = { workspace = true }
-license = { workspace = true }
-repository = { workspace = true }
-version = { workspace = true }
+authors = ["Fuel Labs <contact@fuel.sh>"]
```
pedronauck, line 3: Knope has a bug when using `version = { workspace = true }`, so unfortunately we need to repeat basically what we have in the main `Cargo.toml` in the workspace to be able to have it running properly

0xterminator (approved):

Jurshsmith (approved):

PR #150, opened 2024-08-19
ci(repo): Integrated test coverage framework

0xterminator:


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,76 @@
+name: Coverage
```
pedronauck, line 1: Since this is just one step and it's the same as we already have on `ci.yaml`, you can move it to there instead

pedronauck (changes requested):

.github/workflows/coverage.yaml
0xterminator: of course, done

pedronauck (dismissed):

PR #157, opened 2024-08-23
refactor(data-parser): Make data-parser nats agnostic and more extensibe for benches

Jurshsmith:
Closes DS-58


benchs/bench-consumers/src/runners/runner_all.rs
```diff
@@ -14,13 +13,7 @@ use super::{
...
 pub async fn run_all_benchmarks() -> Result<()> {
     let use_nats_compression = false; // adjust as needed
-    let mut nats = NatsHelper::connect(use_nats_compression).await?;
-    nats.data_parser_mut()
-        .set_serialization_type(SerializationType::Postcard); // adjust as needed
-    nats.data_parser_mut()
-        .set_compression_type(CompressionType::Zlib); // adjust as needed
-    nats.data_parser_mut()
-        .set_compression_level(CompressionLevel::Fastest); // adjust as needed
+    let nats = NatsHelper::connect(use_nats_compression).await?;
```
Jurshsmith, line 16: We can adjust these parameters statically. Please see `connect` fn impl for details

benchs/data-parser/benches/deserialize.rs
```diff
@@ -13,63 +16,49 @@ fn bench_deserialize(c: &mut Criterion) {
...
         .iter()
-        .map(
-            |(serialization_type, compression_type, compression_level)| {
-                let test_block = generate_test_block();
-                let data_parser = DataParserBuilder::new()
-                    .with_compression(*compression_type)
-                    .with_compression_level(*compression_level)
-                    .with_serialization(*serialization_type)
-                    .build();
+        .map(|(serialization_type, compression_strategy)| {
```
Jurshsmith, line 27: `compression_strategy` is sort of a replacement for `compression_type`. 

Also, I removed the `compression_level` for simplicity since we care about the`Fastest` level for the most part. If we are keen on supporting all combinations, we need to define statically them with a suffix like `ZLibFastestCompressionStrategy`, `ZLibBestCompressionStrategy`, et al behind the `bench-helpers` feature flag of course. This might not be ergonomic but it will allow us drastically save a tonne of byte code for our main code.

benchs/nats-publisher/src/utils/tx.rs
```diff
@@ -84,11 +80,7 @@ impl TxHelper {
     ) -> anyhow::Result<()> {
         let tx_id = self.get_id(tx);
         let subject = self.get_subject(tx, block, index);
-        let payload = self
-            .nats
-            .data_parser()
-            .to_nats_payload(&subject.parse(), block)
-            .await?;
```
Jurshsmith, line 91: Simplified encoding and decoding to remove the constraint on some payload structure. Now, `fuel-streams-core` houses the payload structure, `StreamData`, used for streaming which can be encoded and decoded by the `DataParser` with different strategies

pedronauck (approved):

PR #159, opened 2024-08-23
refactor(repo): Added security tests and improvements

0xterminator:


crates/fuel-streams-core/src/stream.rs
```diff
@@ -90,6 +87,7 @@ impl<S: Streamable> Stream<S> {
             .get_or_create_kv_store(kv::Config {
                 bucket: bucket_name.to_owned(),
                 storage: stream::StorageType::File,
+                history: 1,
```
pedronauck, line 90: This property is used to avoid a store entry to have more than one revision at its history

crates/fuel-streams-core/src/stream.rs
```diff
@@ -106,27 +104,36 @@ impl<S: Streamable> Stream<S> {
...
-    ) -> Result<u64, StreamError> {
+    ) -> Result<i64, StreamError> {
         let subject_name = &subject.parse();
-        self.store
-            .put(subject_name, payload.encode(subject_name).await.into())
-            .await
-            .map_err(|s| StreamError::PublishFailed {
+        let result = self
+            .store
+            .create(subject_name, payload.encode(subject_name).await.into())
```
pedronauck, line 111: By using `store.create` instead `store.put` we ensure one subject will be added just once, and we can bypass the `AlreadyExists` to avoid crashing

crates/fuel-streams-core/src/stream.rs
0xterminator: ok, got it.

crates/fuel-streams-core/src/stream.rs
0xterminator: I dont feel this is the cleanest solution, but I agree that its fine to use it for the time being. Having our fuel fork of async_nats should not be an issue in my understanding. @luizstacio what do you reckon here ? Either way, I am fine with the change

crates/fuel-streams-core/src/stream.rs
```diff
@@ -106,27 +104,36 @@ impl<S: Streamable> Stream<S> {
         &self,
         subject: &impl IntoSubject,
         payload: &S,
-    ) -> Result<u64, StreamError> {
+    ) -> Result<i64, StreamError> {
```
Jurshsmith, line 107: Since we don't always have the sequence value, especially in the case when it has been published, it might be better to simplify the interface:

```suggestion
    ) -> Result<(), StreamError> {
```

crates/fuel-streams-core/src/stream.rs
```diff
@@ -106,27 +104,36 @@ impl<S: Streamable> Stream<S> {
...
+        let result = self
+            .store
+            .create(subject_name, payload.encode(subject_name).await.into())
+            .await;
+
+        match result {
+            Ok(sequence) => Ok(sequence as i64),
+            Err(e) if e.kind() == CreateErrorKind::AlreadyExists => {
+                // This is a workaround to avoid publish two times the same message
+                Ok(-1)
```
Jurshsmith, line 118:
```suggestion
            Ok(sequence) => Ok(()),
            Err(e) if e.kind() == CreateErrorKind::AlreadyExists => {
                // This is a workaround to avoid publish two times the same message
                Ok(())
```

crates/fuel-streams-core/src/stream.rs
```diff
@@ -106,27 +104,36 @@ impl<S: Streamable> Stream<S> {
...
         &self,
         // TODO: Allow encapsulating Subject to return wildcard token type
         wildcard: &str,
-    ) -> Result<impl futures::Stream<Item = Vec<u8>>, StreamError> {
-        Ok(self
-            .store
-            .watch(&wildcard)
-            .await
-            .map(|stream| stream.map(|entry| entry.unwrap().value.to_vec()))?)
+    ) -> Result<impl futures::Stream<Item = Option<Vec<u8>>>, StreamError> {
```
Jurshsmith, line 131: Why would we want this?

```suggestion
    ) -> Result<impl futures::Stream<Item = Vec<u8>>, StreamError> {
```

A stream should always yield the published bytes and there should never be a situation where we publish nothing (`None`). What do you think?

crates/fuel-streams-core/src/stream.rs
```diff
@@ -165,28 +172,24 @@ impl<S: Streamable> Stream<S> {
...
         &self,
         wildcard: &str,
     ) -> Result<Option<S>, StreamError> {
         let subject_name = &Self::prefix_filter_subject(wildcard);
-
         let message = self
             .store
             .stream
-            .get_last_raw_message_by_subject(subject_name)
+            .direct_get_last_for_subject(subject_name)
```
Jurshsmith, line 183: Nice 👍🏾. Did we confirm if `allow_direct` is configured to `true`  by default?

pedronauck: It's `true` by default when using kv store
https://github.com/nats-io/nats.rs/blob/b2fb1b406ea9fd2957eadfe2216dea44cd74b56e/async-nats/src/jetstream/context.rs#L731

crates/fuel-streams-core/src/stream.rs
0xterminator: I you check what the entry value is in the closure you will see its a Result with a WatcherErrorKind error. There is two ways we can think here: 1. if error we do an Option and return that not revealing the details to the user. The other option would be to actually have return `Result<impl futures::Stream<Item =Result< Vec<u8>>, WatcherErrorKind>, StreamError>` to the user and let him deal with the error if it occurs. Since we mentioned tha twe dont want to expose the internal details of our impl (here a kv store), we better conceal that error and just map it to an Option. Many other systems do that. The user normally has to do: while let Some(x) = stream.await { ... }. This was my reasoning here. Also, when testing, killing the nats server instantaneously results in WatcherKindError. Previously we were unwrapping it which caused a bad panic. Here is another reason why we want to handle this either 1 or 2 above.

crates/fuel-streams-core/src/stream.rs
0xterminator: good point, done. Ty.

crates/fuel-streams-core/src/stream.rs
0xterminator: agree. Done.

crates/fuel-streams-core/src/stream.rs
Jurshsmith: Thanks for the detailed explanation 👍🏾

PR #208, opened 2024-09-09
feat(publisher): Publish receipts

Jurshsmith:
References:

- https://linear.app/fuel-network/issue/DS-14/receipts-implement-logic-on-publisher
- https://linear.app/fuel-network/issue/DS-21/receipts-implements-subjects-on-core

### DIFF
--- a/crates/fuel-streams-core/src/receipts/subjects.rs
+++ b/crates/fuel-streams-core/src/receipts/subjects.rs
@@ -1,0 +1,136 @@
+use fuel_streams_macros::subject::{IntoSubject, Subject};
+
+use crate::types::*;
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "by_id.receipts.>"]
+#[subject_format = "by_id.receipts.{id_kind}.{id_value}"]
+pub struct ReceiptsByIdSubject {
+    pub id_kind: Option<IdentifierKind>,
+    pub id_value: Option<Bytes32>,
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "receipts.>"]
+#[subject_format = "receipts.{tx_id}.{index}.call.{from}.{to}.{asset_id}"]
+pub struct ReceiptsCallSubject {
+    pub tx_id: Option<Bytes32>,
+    pub index: Option<usize>,
+    pub from: Option<ContractId>,
+    pub to: Option<ContractId>,
+    pub asset_id: Option<AssetId>,
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "receipts.>"]
+#[subject_format = "receipts.{tx_id}.{index}.return.{id}"]
+pub struct ReceiptsReturnSubject {
+    pub tx_id: Option<Bytes32>,
+    pub index: Option<usize>,
+    pub id: Option<ContractId>,
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "receipts.>"]
+#[subject_format = "receipts.{tx_id}.{index}.return_data.{id}"]
+pub struct ReceiptsReturnDataSubject {
+    pub tx_id: Option<Bytes32>,
+    pub index: Option<usize>,
+    pub id: Option<ContractId>,
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "receipts.>"]
+#[subject_format = "receipts.{tx_id}.{index}.panic.{id}"]
+pub struct ReceiptsPanicSubject {
+    pub tx_id: Option<Bytes32>,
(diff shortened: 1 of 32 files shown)


crates/fuel-streams-core/src/receipts/subjects.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "receipts.>"]
+#[subject_format = "receipts.{tx_id}.{index}.panic.{id}"]
+pub struct ReceiptsPanicSubject {
+    pub tx_id: Option<Bytes32>,
+    pub index: Option<usize>,
+    pub id: Option<ContractId>,
+}
```
Jurshsmith, line 49: Repeated similar subjects here to keep it plain simple

crates/fuel-streams-core/src/receipts/subjects.rs
```diff
@@ -0,0 +1,136 @@
+use fuel_streams_macros::subject::{IntoSubject, Subject};
```
Jurshsmith, line 1: I'll have a follow up for documenting these subjects

tests/tests/publisher.rs
```diff
@@ -98,6 +142,57 @@ async fn publishes_transaction_for_each_published_block() {
         .is_ok_and(|result| result.is_some()));
 }
 
+// TODO: Make test more randomized for receipts
```
Jurshsmith, line 145: This test doesn't cover all subject variants for `Receipts` and will be addressed in my follow-up PR

pedronauck (changes requested):

crates/fuel-streams-core/src/types.rs
```diff
@@ -49,6 +70,42 @@ macro_rules! generate_byte_type_wrapper {
...
+/// impl_from_bytes32!(FromType);
+///
+///
+/// Where `FromType` is the type that you want to be able to convert into a `Bytes32`.
+//
+/// # Notes
+///
+/// The macro assumes that the type being converted can be dereferenced into a byte-based type
+/// compatible with the `Bytes32` structure.
+macro_rules! impl_from_bytes32 {
```
0xterminator, line 96: Just a thought but - would not it be better to move this macro to the macros crate (fuel-streams-macros) ?

crates/fuel-streams-macros/src/lib.rs
```diff
@@ -5,17 +5,18 @@ pub mod subject {
...
-                Some(val) => val.to_string(),
-                None => "*".to_string(),
-            }
-        }
-
+    pub trait IntoSubject: std::fmt::Debug + Send + Sync {
         fn parse(&self) -> String;
+    }
+
+    pub trait SubjectBuildable: std::fmt::Debug {
```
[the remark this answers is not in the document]
0xterminator: agree

0xterminator:
Great stuff! Once you resolve the few tiny comments, I believe its good to go!

crates/fuel-streams-macros/src/lib.rs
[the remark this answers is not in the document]
Jurshsmith: > Why do we need another trait here? Isn't it less overhead for our users to import just one trait instead of 2?

https://github.com/FuelLabs/data-systems/blob/be6ff0726e5d17f6bcf7de3879095d1a60b1515a/crates/fuel-streams/src/stream/stream_impl.rs#L19-L20

Yeah, I agree too. It would be really nice to avoid the extra trait. However, the `Filter` interface seems to use `new()`, which isn't object-safe for `IntoSubject`. Hence, the need for the extra trait, `SubjectBuildable`

crates/fuel-streams-core/src/types.rs
Jurshsmith: Good thinking. However, I think the responsibility of `fuel-streams-macros` is to expose procedural macros that, by design, require a `proc-macro` only crate. 

These declarative macros are enhancements for `fuel-core` types which accurately fit into the responsibility of `fuel-streams-core`.

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -121,9 +121,21 @@ impl<S: Streamable> Stream<S> {
         }
     }
 
+    pub async fn publish_many(
```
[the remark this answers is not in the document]
Jurshsmith: Great catch! Done ✅

crates/fuel-streams-macros/src/lib.rs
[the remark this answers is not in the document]
0xterminator: fine by me.

0xterminator (dismissed):

0xterminator (approved):

pedronauck (approved):

PR #211, opened 2024-09-11
feat(publisher): Publish outputs

lostman:
https://linear.app/fuel-network/issue/DS-22/
https://linear.app/fuel-network/issue/DS-15/


crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
+use fuel_core_types::{fuel_tx::TxId, fuel_types::ContractId};
+use fuel_streams_macros::subject::{IntoSubject, Subject};
+
+use crate::types::*;
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "outputs.>"]
+#[subject_format = "outputs.{tx_id}.{index}.{output_type}"]
```
pedronauck, line 8: this isn't the way we have defined in the specs

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
...
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "by_id.outputs.>"]
+#[subject_format = "by_id.outputs.{id_kind}.{id_value}"]
+pub struct OutputsByIdSubject {
+    pub id_kind: Option<IdentifierKind>,
+    pub id_value: Option<Address>,
+}
+
+#[derive(Debug, Clone)]
+pub enum OutputType {
```
pedronauck, line 24: let's try to keep as the standard

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
...
+            OutputType::Change(subject) => subject.parse(),
+            OutputType::Variable(subject) => subject.parse(),
+            OutputType::ContractCreated(subject) => subject.parse(),
+        }
+    }
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "coin.>"]
+#[subject_format = "coin.{from}.{to}.{asset_id}"]
```
pedronauck, line 51: let's try to keep as it is in the tech spec

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl From<CoinSubject> for OutputType {
+    fn from(subject: CoinSubject) -> Self {
+        OutputType::Coin(subject)
+    }
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "contract.>"]
+#[subject_format = "contract.{contract_id}"]
```
pedronauck, line 66: here too

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl From<ContractSubject> for OutputType {
+    fn from(subject: ContractSubject) -> Self {
+        OutputType::Contract(subject)
+    }
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "change.>"]
+#[subject_format = "change.{to}.{asset_id}"]
```
pedronauck, line 79: here too

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl From<ChangeSubject> for OutputType {
+    fn from(subject: ChangeSubject) -> Self {
+        OutputType::Change(subject)
+    }
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "variable.>"]
+#[subject_format = "variable.{to}.{asset_id}"]
```
pedronauck, line 93: here too

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl From<VariableSubject> for OutputType {
+    fn from(subject: VariableSubject) -> Self {
+        OutputType::Variable(subject)
+    }
+}
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "contract_created.>"]
+#[subject_format = "contract_created.{contract_id}"]
```
pedronauck, line 107: here too

pedronauck (changes requested):

crates/fuel-streams-core/src/outputs/subjects.rs
pedronauck: CleanShot 2024-09-12 at 14 59 32@2x

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,143 @@
...
+    OutputsContractSubject::WILDCARD,
+    OutputsChangeSubject::WILDCARD,
+    OutputsVariableSubject::WILDCARD,
+    OutputsContractCreatedSubject::WILDCARD,
+];
+
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "outputs.>"]
+#[subject_format = "outputs.{tx_id}.{index}.>"]
+pub struct OutputsAllSubject {
```
pedronauck, line 19: This one doesn't need to exist since you can do `outputs.>` to get results for all subjects

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,143 @@
+use fuel_core_types::{fuel_tx::TxId, fuel_types::ContractId};
```
pedronauck, line 1: It's missing doc tests here as we have in other subjects

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,143 @@
...
+use fuel_streams_macros::subject::{IntoSubject, Subject};
+
+use crate::types::*;
+
+pub const OUTPUTS_WILDCARD_LIST: &[&str] = &[
```
pedronauck, line 6: Since the only place you have this being consumed is the `impl` in the `mod.rs` you don't need to export this here, you can pass directly into the implementation (as it's in the other places)

0xterminator:
@lostman Please fix the little suggestions by rust here: https://github.com/FuelLabs/data-systems/actions/runs/10889181254/job/30215368841?pr=211 to make the pipeline flash all green

0xterminator (approved):

crates/fuel-streams-core/src/outputs/subjects.rs
pedronauck: But for this query you can do `outputs.*.{tx_id}.{index}.>`, we don't need an extra subject just for this query

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,310 @@
...
+///     .with_tx_id(Some(Bytes32::zeroed()))
+///     .with_index(Some(0));
+/// assert_eq!(
+///     subject.to_string(),
+///     "outputs.0x0000000000000000000000000000000000000000000000000000000000000000.0.>"
+/// );
+/// ```
+#[derive(Subject, Debug, Clone, Default)]
+#[subject_wildcard = "outputs.>"]
+#[subject_format = "outputs.{tx_id}.{index}.>"]
```
pedronauck, line 30: If you check how it works inside the macro, this `subject_format` is being passed for a `format!()` call, so if you do a `>` in the end, you will literally post this `>`

pedronauck (changes requested):

crates/fuel-streams-core/src/outputs/subjects.rs
pedronauck: I think it makes sense to implement this filter in the SDK. Let's proceed with it. 🙏🏻

crates/fuel-streams-core/src/outputs/subjects.rs
pedronauck, line 30:
```suggestion
#[subject_format = "outputs.{tx_id}.{index}"]
```

pedronauck: Hmm? I guess you are miss confusing this here sir

pedronauck: Sorry, sir, I'm not getting your points here. Explain to me why you need this `OutputsAllSubject` in the context of our system. If this is supposed to be a subject to be used inside the `fuel-streams` as a `Filter` or to be published using it in the publisher, this `>` doesn't make sense because you are literally populating `>` in the subject _(in fact, this is the key of the store key)_

As you can [see here](https://github.com/FuelLabs/data-systems/blob/main/crates/fuel-streams-macros/subject-derive/src/parse_fn.rs#L16), this `subject_format` is used to do a parse, this is why the word inside the `{}` need to be the same of the one you passed to the struct, by adding a `>` in the end, you are literally publishing the `>`, this is not used to query.

And, by populating a subject with `>` on it, you are going against what NATS recommends, which is to avoid using `>` literally in the subject.
CleanShot 2024-09-18 at 18 14 08@2x

Do you know what I mean?

crates/fuel-streams-core/src/outputs/subjects.rs
```diff
@@ -0,0 +1,311 @@
+use fuel_core_types::fuel_types::{Address, Bytes32, ContractId};
```
pedronauck, line 1: This is making your tests failing because you are using the wrong types here; you need to use these types, but import from the crate instead
```suggestion
use crate::types::{Address, Bytes32, ContractId};
```

crates/fuel-streams-publisher/src/outputs.rs
```diff
@@ -0,0 +1,135 @@
+use fuel_core_types::fuel_tx::{
```
pedronauck, line 1: If you run `make start` (run docker compose) you will see the publisher is getting errors because also you are using wrong types, instead of importing from `fuel_stream_core` you are importing directly from `fuel_core`

Please, use a `types.rs` file inside the `outputs/` in the `fuel_streams_core` to avoid the need to import `field::Outputs` here and just import from the core as single source

pedronauck (changes requested):

0xterminator (approved):

pedronauck (approved):

PR #223, opened 2024-09-24
feat(repo): Added streaming utxos

0xterminator:


crates/fuel-streams-core/src/types.rs
```diff
@@ -98,6 +98,22 @@ macro_rules! generate_byte_type_wrapper {
...
+        impl AsRef<$inner_type> for $wrapper_type {
+            fn as_ref(&self) -> &$inner_type {
+                &self.0
+            }
+        }
+
+        impl $wrapper_type {
+            pub fn into_inner(self) -> $inner_type {
+                self.0
+            }
```
Jurshsmith, line 116: Nice 👍🏾

crates/fuel-streams-core/src/utxos/types.rs
```diff
@@ -0,0 +1,26 @@
+// ------------------------------------------------------------------------
+// Utxos
+// ------------------------------------------------------------------------
+pub use fuel_core_types::{
+    fuel_tx::{Receipt, Transaction, UniqueIdentifier},
+    services::txpool::TransactionStatus as FuelCoreTransactionStatus,
+};
+
```
Jurshsmith, line 8: The `Utxo` type should probably reside here. wdyt?

crates/fuel-streams-publisher/src/publisher.rs
```diff
@@ -48,12 +74,36 @@ impl Streams {
...
+            ReceiptsReturnSubject::WILDCARD,
+            ReceiptsRevertSubject::WILDCARD,
+            ReceiptsLogDataSubject::WILDCARD,
+            ReceiptsTransferSubject::WILDCARD,
+            ReceiptsMessageOutSubject::WILDCARD,
+            ReceiptsReturnDataSubject::WILDCARD,
+            ReceiptsTransferOutSubject::WILDCARD,
+            ReceiptsScriptResultSubject::WILDCARD,
+            UtxosSubject::WILDCARD,
+            LogsSubject::WILDCARD,
```
Jurshsmith, line 105: Thanks 🙏🏾

crates/fuel-streams-publisher/src/utxos.rs
```diff
@@ -0,0 +1,176 @@
...
+            UtxoType::Message => {
+                publish_with_metrics!(
+                    stream.publish(&subject, &empty_data),
+                    metrics,
+                    chain_id,
+                    block_producer,
+                    UtxosSubject::WILDCARD
+                );
+            }
+        };
```
Jurshsmith, line 172: Sorry, why are we publishing empty Utxos?

crates/fuel-streams-core/src/utxos/mod.rs
```diff
@@ -0,0 +1,16 @@
+pub mod subjects;
+pub mod types;
+
+use serde::{Deserialize, Serialize};
+pub use subjects::*;
+
+use crate::prelude::*;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Utxo(pub Option<Vec<u8>>);
```
Jurshsmith, line 10: This structure seems weird especially since we have some control over creating this data. I don't see why it should be optional

```suggestion
pub struct Utxo(pub Vec<u8>);
```

Jurshsmith, line 10: Nit: Having a small doc for the inner type would also be nice.

Jurshsmith (changes requested):

crates/fuel-streams-publisher/src/utxos.rs
0xterminator: Because I dont know what should go into the Utxo exactly - there is no defined struct Utxo in fuels-core. There is only UtxoId I could send as data @lostman @Jurshsmith @pedronauck  ? ... the ticket on linear does not mention any of that either

crates/fuel-streams-core/src/utxos/mod.rs
0xterminator: yep, but pls see my question/comment above !

0xterminator: yep, will happily add it once we decide on the payload

crates/fuel-streams-publisher/src/publisher.rs
0xterminator: my pleasure

crates/fuel-streams-core/src/utxos/types.rs
0xterminator: agree, but we still need to decide if it should be a self defined sturct or we could get it somewhere from fuels-core which I could not find

crates/fuel-streams-core/src/types.rs
0xterminator: my pleasure

crates/fuel-streams-core/src/utxos/mod.rs
0xterminator: added utxoId

crates/fuel-streams-core/src/utxos/subjects.rs
```diff
@@ -0,0 +1,191 @@
...
+#[allow(clippy::too_many_arguments)]
+pub struct UtxosSubject {
+    pub hash: Option<MessageId>,
+    pub tx_id: Option<Bytes32>,
+    pub utxo_type: Option<UtxoType>,
+    pub sender: Option<Address>,
+    pub recipient: Option<Address>,
+    pub nonce: Option<Nonce>,
+    pub data: Option<String>, // hexified data
+    pub amount: Option<u64>,
```
Jurshsmith, line 95: I think these are meant to be in the `Utxo` struct, not here in the subject

0xterminator: yes, good point. I have point. I have ported it over. Ty

crates/fuel-streams-publisher/src/utxos.rs
```diff
@@ -0,0 +1,204 @@
...
+                publish_with_metrics!(
+                    stream.publish(&subject, &utxo),
+                    metrics,
+                    chain_id,
+                    block_producer,
+                    UtxosSubject::WILDCARD
+                );
+            }
+        };
+    }
```
Jurshsmith, line 201:
```suggestion
    for (subject, utxo) in subjects {
        publish_with_metrics!(
              stream.publish(&subject, &utxo),
              metrics,
              chain_id,
              block_producer,
              UtxosSubject::WILDCARD
          );
     }
```

crates/fuel-streams-core/src/utxos/types.rs
```diff
@@ -0,0 +1,95 @@
...
+};
+// ------------------------------------------------------------------------
+// Utxos
+// ------------------------------------------------------------------------
+pub use fuel_core_types::{
+    fuel_tx::{Receipt, Transaction, UniqueIdentifier},
+    services::txpool::TransactionStatus as FuelCoreTransactionStatus,
+};
+use serde::{Deserialize, Serialize};
+// use super::{types::UtxoType, Address, Bytes32, MessageId, Nonce};
```
Jurshsmith, line 19: Do we still need these?

crates/fuel-streams-core/src/utxos/types.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Utxo {
+    pub id: UtxoId,
+    pub sender: Option<Address>,
+    pub recipient: Option<Address>,
+    pub nonce: Option<Nonce>,
+    pub data: Option<Vec<u8>>,
+    pub amount: Option<u64>,
+    pub tx_id: Option<Bytes32>,
```
Jurshsmith, line 29: Uxos originate from transactions, so this can't be optional:

```suggestion
    pub tx_id: Bytes32,
```

Jurshsmith (changes requested):

crates/fuel-streams-core/src/utxos/types.rs
0xterminator: true, fixed

crates/fuel-streams-core/src/utxos/types.rs
0xterminator: no, removed them.

crates/fuel-streams-publisher/src/utxos.rs
0xterminator: yep. true :) Thanks.

Jurshsmith (approved):

pedronauck (approved):

PR #224, opened 2024-09-26
docs(publisher): Add an example of streaming contract events. DS-64

lostman:


0xterminator (dismissed):
Good stuff!

examples/streams/src/main.rs
```diff
@@ -124,3 +151,83 @@ async fn stream_transactions(
...
+    contract_id: ContractId,
+) -> anyhow::Result<()> {
+    let mut receipt_stream = fuel_streams::Stream::<Receipt>::new(client).await;
+
+    receipt_stream.with_filter(
+        ReceiptsBurnSubject::new().with_contract_id(Some(contract_id.into())),
+    );
+    receipt_stream.with_filter(
+        ReceiptsCallSubject::new().with_from(Some(contract_id.into())),
+    );
```
Jurshsmith, line 184: This might be better from a readability standpoint:

```suggestion
    let mut receipt_stream = fuel_streams::Stream::<Receipt>::new(client).await;
    let receipt_stream = receipt_stream.with_filter(
        ReceiptsBurnSubject::new().with_contract_id(Some(contract_id.into())),
    ).with_filter(
        ReceiptsCallSubject::new().with_from(Some(contract_id.into())),
    ).
    // other with_filters come here
```

pedronauck (commented):
How about transactions and other resources from a specific contract id? You can use the `by_id` subject to add examples too.

examples/publisher/Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
-name = "fuel-streams-publisher-examples"
+name = "fuels-example-publisher"
```
Jurshsmith, line 2: Sorry, this is a bit confusing. Why this change? Are the examples for `fuels` or `fuel-streams`?

examples/streams/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "fuel-streams-examples"
+name = "fuels-example-streams"
```
Jurshsmith, line 2: same here

pedronauck (dismissed):

PR #240, opened 2024-09-27
feat(repo): Parallel publishing of multiple streams (optimization)

0xterminator:
This is a small PR which boosts up the multithreaded paralellized sending of multiple streams. Locally on my machine it does lead to a performance boost.


Jurshsmith (approved):
Good stuff 👍🏾

pedronauck (approved):

PR #245, opened 2024-10-06
feat(repo): Added fuel streams protobuf crate and extended data-parser

0xterminator:


crates/fuel-streams-protobuf/proto/fuel_types.proto
```diff
@@ -0,0 +1,20 @@
...
+message BlockHeader {
+    uint32 da_height = 1;
+    uint32 consensus_parameters_version = 2;
+    uint32 state_transition_bytecode_version = 3;
+}
+
+message Transaction {
+    bytes mint_asset_id = 1;
+    uint64 mint_amount = 2;
+}
```
0xterminator, line 20: I am just making a note here: This structure is in reality very nested with a lot of variants for Mint, Script etc. - highly unpractical for mapping to protobuf as the latter does not support what Rust does.

pedronauck:
Let's keep this as draft until we decide about the parser

PR #246, opened 2024-10-06
feat(repo): Added codespell to gh actions and Makefile

0xterminator:
This MR adds codespell locally and to gh actions


pedronauck (changes requested):
Why do we need this if we already have `.typos.yaml` installed?

0xterminator:
We can close it for now!

PR #261, opened 2024-10-15
refactor(publisher): Use identifiers for scripts and predicates

Jurshsmith:
References: DS-89

### DIFF
--- a/crates/fuel-streams-publisher/src/inputs.rs
+++ b/crates/fuel-streams-publisher/src/inputs.rs
@@ -12,18 +12,7 @@ use fuel_core_types::fuel_tx::{
     },
     UniqueIdentifier,
 };
-use fuel_streams_core::{
-    inputs::{
-        InputsByIdSubject,
-        InputsCoinSubject,
-        InputsContractSubject,
-        InputsMessageSubject,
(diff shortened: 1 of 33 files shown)


0xterminator (approved):

crates/fuel-streams-publisher/src/inputs.rs
```diff
@@ -20,239 +25,129 @@ use fuel_streams_core::{
...
+                            InputsMessageSubject::new()
+                                .with_tx_id(Some(tx_id.into()))
+                                .with_index(Some(index))
+                                .with_sender(Some(sender.into()))
+                                .with_recipient(Some(recipient.into()))
+                                .boxed(),
+                            InputsMessageSubject::WILDCARD,
+                        ),
+                        (
+                            InputsByIdSubject::new()
```
pedronauck, line 126: I guess we can publish two times here, one for `sender` and another for `recipient`, what do you think?

pedronauck (changes requested):

pedronauck (commented):
Nice job 👏🏻

pedronauck (approved):

PR #263, opened 2024-10-15
fix(publisher): Publish transactions by id

lostman:
We are publishing transactions, but not transactions by ID. This PR fixes that.


0xterminator (approved):

pedronauck (changes requested):
I guess doesn't make sense have a transaction ID here as IdentifierKind, because you can already get a transaction by id using `transactions.*.*.{id}.>` what we need on this PR is to filter transactions by address, contract ID and by asset ID and use start publishing the `TransactionByIdSubject` within the publisher

pedronauck:
I'm get this task to finish

PR #264, opened 2024-10-15
refactor(publisher): Refactor examples

lostman:
This PR refactors the examples:

1. Removed client and publisher examples, as this is something other than what the users would do.
2. Renamed streams to multiple-streams
3. Examples now work with `cargo run --example transactions/receipts/multiple-streams`


examples/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "fuel-data-systems-examples"
```
pedronauck, line 2:
```suggestion
name = "fuel-streams-examples"
```

examples/receipts.rs
```diff
@@ -0,0 +1,123 @@
...
+///
+/// # Arguments
+///
+/// * `client` - A reference to the NATS client used for streaming.
+/// * `contract_id` - The ID of the contract to monitor.
+///
+/// # Returns
+///
+/// Returns `Ok(())` if the streaming completes successfully, or an error if there are any issues.
+async fn stream_contract_events(
```
pedronauck, line 43: please, add everything in a `main` function instead of call this other function, as much simple as possible

pedronauck (changes requested):

Jurshsmith (dismissed):

0xterminator:
@lostman I dont understand why we need separate examples for receipts e.g. provided they are covered by the multiple-streams. My original idea was to put everything in the streams so that the user could just check what we offer and use it. I fyou do that, you would also have to create N other examples for logs, utxos etc. I dont agree we dont need the client ones just because you as a user would have to build a client first ot be able to use the streams. We just demonstrate the different abilities of building a client. I think you might want to keep the publisher too as this would simply show how the sdk could also be used in  a different context in case the user would want to publish messages to his own context e.g. I find it useful as many users would want to do something else with the fuel messages they read.

pedronauck:
> @lostman I dont understand why we need separate examples for receipts e.g. provided they are covered by the multiple-streams. My original idea was to put everything in the streams so that the user could just check what we offer and use it. I fyou do that, you would also have to create N other examples for logs, utxos etc. I dont agree we dont need the client ones just because you as a user would have to build a client first ot be able to use the streams. We just demonstrate the different abilities of building a client. I think you might want to keep the publisher too as this would simply show how the sdk could also be used in a different context in case the user would want to publish messages to his own context e.g. I find it useful as many users would want to do something else with the fuel messages they read.

@0xterminator, we discussed this in a meeting that you couldn't participate in. This examples folder is something very focused on the user that consumes the SDK; it's a place they will look to get an overview of how the code works and figure out how to use the project. They need to be simple, and the user needs to quickly find what they want; adding everything inside a single `stream.rs` file makes it harder to find out. I showed up [this](https://github.com/nats-io/nats.rs/tree/main/async-nats/examples) and [this](https://github.com/FuelLabs/fuels-rs/tree/master/examples) examples folder as a reference.

But I agree with you. Since @lostman separated transactions and receipts, he should, of course, add the rest.

pedronauck (changes requested):
Can you add the rest of the modules as example?

Jurshsmith (approved):

PR #277, opened 2024-10-22
refactor(data-parser): Serialize in json by default

Jurshsmith:
Closes DS-95


0xterminator (approved):

pedronauck (changes requested):
@Jurshsmith can you check about using the `.json()` method that fuel core exposes for block and transactions? if it's possible to use them to format data and also if it's other modules (inputs, outputs, receipts, etc) has the same method... because by just serializing for JSON we will get a very weird data format that is the one we are consuming right now

Jurshsmith:
@pedronauck as shown in the screenshot below, the `to_json` and `from_json` implementations use `serde` under the hood which produces the our current output. 

Is there any particular issue with the currently outputted JSON? Would we prefer to serialize to JSON string instead of a raw bytes (Vec<u8>)?
 
Screenshot 2024-10-25 at 20 01 55

PR #282, opened 2024-10-24
fix(publisher): Missing by id subjects 

pedronauck:
## Tasks

- [x] Fix all `ByIdSubjects` inside the publisher
  - [x] Missing subjects in some cases within inputs, outputs and receipts
  - [x] Publish the proper subjects for transactions (`TransactionsByIdSubjects`)
- [x] Improve parallelism/concurrency when publishing streams

## Improvements

- By using `rayon`, `futures` and `tokio` to do proper parallelism and concurrency in the publisher, it seems that we got very nice improvement in terms of speed ⚡ 
CleanShot 2024-10-25 at 00 00 35@2x
- Now, the average time to sync 1 million blocks from Mainnet is around 10min (with a populated db)
CleanShot 2024-10-25 at 00 03 28


0xterminator (approved):
Works fine by me! We can surely refine it in a later phase, but the re-work is quite meaningful at this stage!

PR #283, opened 2024-10-24
fix(core): Added retry in case of store creation timeout

0xterminator:
Added retry strategy for store creation in case of a timeout  (currently happening on the cluster).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3555,6 +3555,7 @@ dependencies = [
  "serde",
  "thiserror",
  "tokio",
+ "tokio-retry",
  "tracing",
 ]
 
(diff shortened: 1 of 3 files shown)


pedronauck (dismissed):

Jurshsmith (changes requested):
Let's delegate retries to Kubernetes for now. The store creation is failing due to NATS' previous OOM error. The cluster is in a state where it can't handle incoming traffic because it is migrating all the old data from the 400 GiB memory to the new 800 GiB

pedronauck:
> Let's delegate retries to Kubernetes for now. The store creation is failing due to NATS' previous OOM error. The cluster is in a state where it can't handle incoming traffic because it is migrating all the old data from the 400 GiB memory to the new 800 GiB

I agree with AJ on that, I don't issues when creating stores to add a retry mechanism right now

PR #284, opened 2024-10-24
feat(publisher): Cleanup the return type of Publisher::run()

lostman:
```
pub async fn run(mut self) -> anyhow::Result<Self>
```

is rather odd. Internally, it loops forever (until shutdown). It is confusing that it returns `Self` (i.e. `Publisher`).

Normally, such functions return `()`.

```
let publisher = publisher.run().await.unwrap();
```

This is also an anti-pattern.

This PR changes this and replaces the uses as shown above with more idiomatic Rust.

### DIFF
--- a/crates/fuel-streams-publisher/src/publisher.rs
+++ b/crates/fuel-streams-publisher/src/publisher.rs
@@ -227,7 +227,7 @@ impl Publisher {
         Ok(())
     }
 
-    pub async fn run(mut self) -> anyhow::Result<Self> {
+    pub async fn run(&mut self) -> anyhow::Result<()> {
         let mut stop_handle = StopHandle::new();
         stop_handle.spawn_signal_listener();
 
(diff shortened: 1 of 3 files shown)


crates/fuel-streams-publisher/src/publisher.rs
```diff
@@ -227,7 +227,7 @@ impl Publisher {
         Ok(())
     }
 
-    pub async fn run(mut self) -> anyhow::Result<Self> {
+    pub async fn run(&mut self) -> anyhow::Result<()> {
```
0xterminator, line 230: @lostman I think the idea here is (and this is not my code btw.) that once you create an instance of the Publisher, you can call .run() on it and it would consume self meaning you would not be allowed to run any other methods once you call run on it. I personally prefer this design. Of course returning self is not great here, so you changes to () are fine. Nevertheless ,I feel the signatrure argument should be as before pub async fn run(mut self)

0xterminator: @Jurshsmith @pedronauck  ?

Jurshsmith: Exactly 👍🏾 @0xterminator. The old signature explicitly consumes(or moves) the publisher to remove the ambiguity of reactive mutations after running.

PR #303, opened 2024-11-08
ci(repo): Fix test action

pedronauck:


.github/workflows/ci.yaml
```diff
@@ -23,149 +23,149 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  validate-title:
```
0xterminator, line 26: Dont we want to have these steps anymore ?

PR #310, opened 2024-11-12
feat(repo): Restructured publisher into smaller modules

0xterminator:


crates/fuel-streams-publisher/src/payloads/mod.rs
```diff
@@ -0,0 +1,8 @@
+pub mod blocks;
+pub mod identifiers;
+pub mod inputs;
+pub mod logs;
+pub mod outputs;
+pub mod receipts;
+pub mod transactions;
+pub mod utxos;
```
Jurshsmith, line 8: Since only the publisher consumes these, this should reside in the `publisher` module

Jurshsmith: As a sub-sub module of course

pedronauck (dismissed):

Jurshsmith (approved):

pedronauck (approved):

PR #312, opened 2024-11-14
refactor(publisher): remove `PublishError` in favor of `anyhow::Result`

Jurshsmith:
This commit removes PublishError in favor of bubbling pending unactionable errors to std when there is an error publishing any subject/payload.

Given the focus of the Publish process, only dependent libraries or actionable errors should have an error signature. We can safely do this since it's all application code and we will be relying on our tracing for stack traces to implement certain actions after analyzing the errors. This makes our code more concise for the domain solution it solves.

Closes DS-108


0xterminator (approved):
NIce. WOnder if we should also add .context("xxx") from the anyhow crate for extra info to some of the errors, or maybe not ?

PR #315, opened 2024-11-19
feat(repo): Added local websocket for nats

0xterminator:


Jurshsmith (dismissed):

ws-test.html
```diff
@@ -0,0 +1,32 @@
+<!doctype html>
```
pedronauck, line 1: Can you add this inside the `examples` folder instead?

pedronauck (changes requested):

PR #320, opened 2024-11-23
feat(repo): Introduce FuelStreams Types

Jurshsmith:
Currently, `fuel-core` types are directly exposed to users for streaming. However, this approach has several drawbacks:

1. The serialized structures are tailored for Rust, making them difficult to integrate with existing SDKs.
2. The unstructured serialized output complicates compatibility efforts.
3. Exposing `fuel-core` types could mislead users into assuming related APIs, which require a configured environment, are usable directly with streams.

This PR addresses these issues by introducing `FuelStreams` types, designed as aggregation types optimized for broad use cases. These types are derived from the GraphQL interface used in frontend projects, ensuring a structured and user-friendly format.

Moving forward, we will expand these types to include additional fields based on user needs, making them more versatile and compatible across different integrations.

Closes DS-113

_Subtasks:_
Closes DS-114, DS-115, DS-116, DS-117, DS-118


pedronauck (approved):

crates/fuel-streams-core/src/transactions/types.rs
```diff
@@ -1,27 +1,426 @@
...
+                    Some(mint.input_contract().into())
+                }
+                _ => None,
+            }
+        };
+
+        let input_contracts = {
+            use std::vec::IntoIter;
+
+            fn input_contracts<Tx>(
```
pedronauck, line 125: This function is being created on each new `Transaction::new()` execution; since it's a pure function, I guess you could add it as a `Transaction` method or get it out of the this scope

pedronauck: And if you just do this is not the same?

```rust
        let input_contracts = {
            match transaction {
                FuelCoreTransaction::Mint(_) => None,
                tx => {
                    let mut inputs: Vec<_> = tx
                        .inputs()
                        .iter()
                        .filter_map(|input| match input {
                            fuel_tx::Input::Contract(contract) => {
                                Some(contract.contract_id)
                            }
                            _ => None,
                        })
                        .collect();
                    inputs.sort();
                    inputs.dedup();
                    Some(inputs.into_iter().map(|id| (*id).into()).collect())
                }
            }
        };
```

Jurshsmith: Nice 🚀. Will update.

Jurshsmith: Done ✅

pedronauck (approved):

0xterminator (approved):

PR #338, opened 2024-12-02
build(repo): Added load tester

0xterminator:


benches/load-tester/src/runners/runner_blocks.rs
```diff
@@ -0,0 +1,51 @@
...
+    let mut sub = stream.subscribe_with_config(config).await?;
+
+    // Process incoming blocks
+    while let Some(bytes) = sub.next().await {
+        match bytes.as_ref() {
+            Err(_) => load_test_tracker.increment_error_count(),
+            Ok(message) => {
+                load_test_tracker.increment_message_count();
+                let decoded_msg =
+                    Block::decode_raw(message.payload.to_vec()).await;
```
Jurshsmith, line 33: Decoding the actual type shouldn't be a concern in these load benches

Jurshsmith (changes requested):

benches/load-tester/src/runners/runner_blocks.rs
0xterminator: The only reason I am decoding it back is to retrieve the timestamp of publishing of the StreamData as I am later doing some stats based on that.

0xterminator: It has no effect on the load testing itself.

benches/load-tester/src/runners/runner_all.rs
```diff
@@ -0,0 +1,239 @@
...
+        // logs
+        let logs_test_tracker =
+            Arc::new(LoadTestTracker::new("Logs Consumer".into()));
+        let logs_test_tracker_printer = Arc::clone(&logs_test_tracker);
+
+        // outputs
+        let outputs_test_tracker =
+            Arc::new(LoadTestTracker::new("Outputs Consumer".into()));
+        let outputs_test_tracker_printer = Arc::clone(&outputs_test_tracker);
+
```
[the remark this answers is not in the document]
0xterminator: The idea was here to apply all sorts of streaming to the cluster and not just blocks or txs. In fact logs and e.g. utxos are very small against blocks and txs so I presume  their contrib is quite small.

benches/load-tester/Cargo.lock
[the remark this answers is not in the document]
0xterminator: you mean part of the global workspace and not the benches ? The load-tester is not a bench so it might make sense to move it to the global workspace if that is what you mean ?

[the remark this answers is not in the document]
Jurshsmith: it can stay in the `/benches`, but no need for an extra `Cargo.lock` since all the dependencies can be gotten from the workspace's root

benches/load-tester/src/runners/results.rs
```diff
@@ -0,0 +1,118 @@
...
+        // Lock the mutex to access publish_times
+        let times = self.publish_times.lock().unwrap();
+
+        if times.is_empty() {
+            return;
+        }
+
+        let times_ns: Vec<f64> =
+            times.iter().map(|d| d.as_nanos() as f64).collect();
+        drop(times);
```
Jurshsmith, line 109: Nit:

```suggestion
            times.into_iter().map(|d| d.as_nanos() as f64).collect();
```

benches/load-tester/Cargo.lock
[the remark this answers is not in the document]
0xterminator: yep, understood. I have removed the cargo.locks

benches/load-tester/src/runners/runner_blocks.rs
Jurshsmith: Okay. But it would be nice to have a single runner that accepts a `impl Streamable` stream argument to dynamically compute these. Repeating the same logic per stream type makes it seem as though the load testing has some biases based on the stream type

0xterminator: done

benches/load-tester/src/runners/results.rs
0xterminator: cannot do into_iter as we are under a Mutex and we cannot consume it here.

Jurshsmith (approved):

pedronauck (approved):

PR #340, opened 2024-12-03
fix(publisher): Ensure OffchainDB is at least one block ahead of the currently published block

Jurshsmith:
References: DS-124


0xterminator (approved):

pedronauck:
I'll close this for now since we are already working in other solution for this @Jurshsmith

PR #348, opened 2024-12-06
feat(publisher): Extend NATS publishing by storing payloads in S3

Jurshsmith:
This PR enhances our current Publisher strategy by publishing 
the payloads in S3 and storing the path pointer in NATS

Closes DS-125


crates/fuel-streams-storage/src/s3/s3_client_opts.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    pub fn endpoint_url(&self) -> String {
+        match self.role {
+            FuelNetworkUserRole::Admin => dotenvy::var("AWS_S3_ENDPOINT_URL")
+                .expect("AWS_S3_ENDPOINT_URL must be set for admin role"),
+            FuelNetworkUserRole::Default => self.fuel_network.to_s3_url(),
+        }
+    }
+
+    pub fn region(&self) -> String {
```
mchristopher, line 50: This shouldn't be manual - the AWS SDK will handle it by default if you don't set it.

crates/fuel-streams-storage/src/s3/s3_client_opts.rs
```diff
@@ -0,0 +1,71 @@
...
+        self.namespace = Some(random_namespace);
+        self
+    }
+
+    pub fn with_role(self, role: FuelNetworkUserRole) -> Self {
+        Self { role, ..self }
+    }
+
+    pub fn endpoint_url(&self) -> String {
+        match self.role {
```
mchristopher, line 43: This should be optional - if it's not provided, don't set it. We should never set this manually when we're deploying, it's only useful for tests or when running outside of AWS.

Jurshsmith: Right. This code also runs outside AWS for client apps. But I'll make this the default `AWS_ENDPOINT_URL` and optionally set it as you've recommended.

0xterminator (approved):

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -65,13 +65,27 @@ pub trait Streamable: StreamEncoder {
     fn to_packet(&self, subject: Arc<dyn IntoSubject>) -> PublishPacket<Self> {
         PublishPacket::new(self.clone(), subject)
     }
+
+    fn get_s3_path(&self) -> String {
+        format!("v1/{}/{}.json", Self::NAME, self.get_consistent_hash())
```
pedronauck, line 70: Since we are using the `nats_client.namespace` to create the name for the bucket (`fuel_<bucket>`) would be not nice to have the namespace here also to create the path?

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -199,28 +224,101 @@ impl<S: Streamable> Stream<S> {
...
 
-    // TODO: This should probably be `subscribe_raw` since it returns pure bytes
+    // Less performant due to our hybrid use of NATS and S3
+    pub async fn subscribe_raw(
+        &self,
+        subscription_config: Option<SubscriptionConfig>,
+    ) -> Result<BoxStream<'_, Vec<u8>>, StreamError> {
+        let mut config = PullConsumerConfig {
+            filter_subjects: vec![S::WILDCARD_LIST[0].to_string()],
+            deliver_policy: DeliverPolicy::All,
```
pedronauck, line 234:
```suggestion
            filter_subjects: subscription_config.unwrap_or(vec![S::WILDCARD_LIST[0].to_string()]),
            deliver_policy: subscription_config.unwrap_or(DeliverPolicy::All),
```

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -199,28 +224,101 @@ impl<S: Streamable> Stream<S> {
...
+            ack_policy: AckPolicy::None,
+            ..Default::default()
+        };
+
+        if let Some(subscription_config) = subscription_config {
+            config.filter_subjects = subscription_config.filter_subjects;
+            config.deliver_policy = subscription_config.deliver_policy;
+        }
```
pedronauck, line 242:
```suggestion
```

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -199,28 +224,101 @@ impl<S: Streamable> Stream<S> {
...
+
+        Ok(consumer
+            .messages()
+            .await?
+            .then(|message| {
+                let s3_client = Arc::clone(&self.s3_client);
+
+                async move {
+                    let nats_payload = message
+                        .expect("Message must be valid")
```
pedronauck, line 255: My main concern here (and in other parts of the project as well), is that we are panicking a lot 🙄 IDK, I guess we should handle these errors better to prevent programs from crash entirely

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -199,28 +224,101 @@ impl<S: Streamable> Stream<S> {
...
-        wildcard: &str,
-    ) -> Result<impl futures::Stream<Item = Option<Vec<u8>>>, StreamError> {
-        Ok(self.store.watch(&wildcard).await.map(|stream| {
-            stream.map(|entry| {
-                entry.ok().map(|entry_item| entry_item.value.to_vec())
+        subscription_config: Option<SubscriptionConfig>,
+    ) -> Result<BoxStream<'_, S>, StreamError> {
+        let mut config = PullConsumerConfig {
+            filter_subjects: vec![S::WILDCARD_LIST[0].to_string()],
+            deliver_policy: DeliverPolicy::All,
```
pedronauck, line 278:
```suggestion
            filter_subjects: subscription_config.unwrap_or(vec![S::WILDCARD_LIST[0].to_string()]),
            deliver_policy: subscription_config.unwrap_or(DeliverPolicy::All),
```

crates/fuel-streams-core/src/stream/stream_impl.rs
```diff
@@ -199,28 +224,101 @@ impl<S: Streamable> Stream<S> {
...
+            ack_policy: AckPolicy::None,
+            ..Default::default()
+        };
+
+        if let Some(subscription_config) = subscription_config {
+            config.filter_subjects = subscription_config.filter_subjects;
+            config.deliver_policy = subscription_config.deliver_policy;
+        }
```
pedronauck, line 286:
```suggestion
```

crates/fuel-networks/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+impl FuelNetwork {
+    pub fn load_from_env() -> Self {
+        match std::env::var("NETWORK").as_deref() {
+            Ok("testnet") => FuelNetwork::Testnet,
+            Ok("mainnet") => FuelNetwork::Mainnet,
+            _ => FuelNetwork::Local,
+        }
+    }
+
+    pub fn to_nats_url(&self) -> String {
```
pedronauck, line 37: By pinning the NATS URL here, we will have the same issue we had in the past: not being able to pass a custom `NATS_URL` to internal cluster traffic. As mentioned previously, when running this internally in the cluster, instead of going directly to the NATS node host `nats://nats.<namespace>.svc.cluster.local:4222`, we will always go through the external service NLB.

crates/fuel-streams-publisher/src/cli.rs
```diff
@@ -9,15 +9,6 @@ use clap::Parser;
 /// - `fuel_core_config`: Configuration for the Fuel Core service, parsed using a flattened command.
 #[derive(Clone, Parser)]
 pub struct Cli {
-    /// Fuel Network to connect to.
-    #[arg(
```
pedronauck, line 13: We should remove this, as mentioned previously in other comment.

crates/fuel-streams-storage/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "fuel-streams-storage"
```
pedronauck, line 2: With the actual content of this package, I don't know if we should call it `fuel-streams-storage`. We could have a package with this name that uses the Adapter pattern to handle different types of sources (not only S3), but at the moment, this package just has two clients (S3 client and NATS client). So, for now, since we don't have the Adapter pattern, I don't see why we should have this new package; both clients could stay in the `fuel-streams-core` for now, IMHO.

crates/fuel-streams/src/stream/stream_impl.rs
```diff
@@ -185,15 +228,64 @@ impl<S: Streamable> Stream<S> {
...
+    /// use fuel_streams::blocks::Block;
+    /// use fuel_streams::types::DeliverPolicy;
+    ///
+    /// # async fn example() -> Result<(), fuel_streams::Error> {
+    /// # let client = Client::connect(FuelNetwork::Local).await?;
+    /// # let stream = Stream::<Block>::new(&client).await;
+    /// let config = StreamConfig {
+    ///     deliver_policy: DeliverPolicy::All,
+    /// };
+    /// let subscription = stream.subscribe_with_config(config).await?;
```
pedronauck, line 272:
```suggestion
    /// let subscription = stream.subscribe_raw_with_config(config).await?;
```

crates/fuel-streams/src/stream/stream_impl.rs
```diff
@@ -185,15 +228,64 @@ impl<S: Streamable> Stream<S> {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn subscribe_raw_with_config<'a>(
```
pedronauck, line 276: We should consider whether external users really need to subscribe for raw bytes. For now, it might be better to keep this method internal to avoid bloating the SDK. It could be better to decode internally in the SDK and deliver the final typed data instead.

examples/blocks.rs
```diff
@@ -31,14 +31,13 @@ async fn main() -> anyhow::Result<()> {
...
-    let mut sub = stream.subscribe_with_config(config).await?;
+    let mut sub = stream.subscribe_raw_with_config(config).await?;
 
     println!("Listening for blocks...");
 
     // Process incoming blocks
     while let Some(bytes) = sub.next().await {
-        let message = bytes?;
-        let decoded_msg = Block::decode_raw(message.payload.to_vec()).await;
+        let decoded_msg = Block::decode_raw(bytes).unwrap();
```
pedronauck, line 40: If we choose to decode the data internally in the SDK, we could get rid of this and avoid one extra step here, improving DX.

Jurshsmith: Yeah, that's the plan. However, we have to make that change progressively. For instance, the benches, load tests, and many other tests need the raw bytes atm. Definitely a post-s3 publishing task.

crates/fuel-networks/src/lib.rs
Jurshsmith: Please see: https://github.com/FuelLabs/data-systems/blob/d554ed1474208015a666173cfd849bc28bd30308/crates/fuel-streams-storage/src/nats/nats_client_opts.rs#L70,

That's where we use the env variable for the admin role. For default roles, we will be using these hardcoded values

crates/fuel-streams-storage/Cargo.toml
Jurshsmith: The core module is expanding significantly, leading to increased coupling between components. To address this, we should progressively break it down into smaller, focused crates. This approach will not only improve build speed but also reduce coupling, making the codebase easier to maintain and scale.

From an architectural perspective, smaller crates will provide better visibility into how major components are interconnected, improving overall maintainability and enabling clearer boundaries between concerns.

As a first step, we propose encapsulating storage-related logic into a dedicated crate. This will explicitly define how we extend functionality with NATS, S3, and other storage systems. Shared modules like Namespace and Role will remain accessible, but logically grouped to streamline reuse and clarity.

Over time, we should envision further refining this approach by breaking the storage crate into more granular internal crates, such as nats and s3. These crates will encapsulate their respective logic and be used only where needed, fostering a modular and maintainable system design.

crates/fuel-streams-core/src/stream/stream_impl.rs
Jurshsmith: I am not sure what that helps with. The namespace boundary is already achieved at the bucket level. Repeating it here would be redundant, right?

crates/fuel-streams-core/src/stream/stream_impl.rs
Jurshsmith: Here we `unwrap` twice, and as the config grows we will keep unwrapping. Mutating after the default would be more maintainable here. 

Also, most of these user-exposed functions here need some re-visiting. we shouldn't have these many `subscribe` functions but since that's the current state, I'd suggest we tackle those in a well-refined task

crates/fuel-streams-core/src/stream/stream_impl.rs
Jurshsmith: Should we let the user handle unexpected errors at this early stage? In this case, we always publish with a value, our program ensures that, so the `expect` there is an assertion that `async_nats` couldn't help us do. The maintainability of our code improves greatly by identifying these sorts of guarantees

Jurshsmith (commented):
> I'm missing also the configuration to have this running in the local cluster, should we have this configured on this PR or then?

Did you update your `.env` with the new `.env.sample` variables?

crates/fuel-streams-core/src/stream/stream_impl.rs
pedronauck: As you wish 🙏🏻

crates/fuel-streams-core/src/stream/stream_impl.rs
pedronauck: I understand your perspective, and I see the benefit of using `expect()` to show invariants clearly. However, I worry that depending on panics for minor issues might hurt DX and operational stability. Given that we’re running a streaming service 24/7, even one unexpected value could crash the app, halt the stream, and affect users. This isn't good.

How about we find a middle ground? We could use `expect()` only in critical areas while implementing proper error handling and logging where problems could be eliminated. This way, we keep the guarantees without risking major issues from minor glitches.

crates/fuel-streams-storage/Cargo.toml
pedronauck: I'm not disagreeing with you; I'm just saying that now isn't the time to do that. We are rushing to deliver. We could sacrifice some "good practices" and architectural decisions for now in order to be faster. But ok, if this didn't slowing down you, np 🙏🏻

pedronauck:
@Jurshsmith 
> Did you update your .env with the new .env.sample variables?

I mean, did you test this using the Minikube and local Cluster, it's working right away with it?

PR #349, opened 2024-12-09
refactor(repo): Improve DevOps stack and configuration

pedronauck:


crates/fuel-streams-publisher/src/publisher/fuel_core_like.rs
```diff
@@ -34,7 +34,10 @@ pub type OffchainDatabase = GenericDatabase<
 pub trait FuelCoreLike: Sync + Send {
     async fn start(&self) -> anyhow::Result<()>;
     fn is_started(&self) -> bool;
-    async fn await_synced_at_least_once(&self) -> anyhow::Result<()>;
+    async fn await_synced_at_least_once(
+        &self,
+        historical: bool,
```
Jurshsmith, line 39: This is a bit of an abstraction leak. `fuel-core` doesn't have a `historical` domain, it's more of the `publisher`'s concern

crates/fuel-streams-publisher/src/publisher/mod.rs
```diff
@@ -103,10 +103,13 @@ impl Publisher {
...
         &self,
         shutdown_token: ShutdownToken,
+        historical: bool,
     ) -> anyhow::Result<()> {
         tracing::info!("Awaiting FuelCore Sync...");
 
-        self.fuel_core.await_synced_at_least_once().await?;
+        self.fuel_core
+            .await_synced_at_least_once(historical)
+            .await?;
```
Jurshsmith, line 112:
```suggestion
if !historical {
        self.fuel_core
            .await_synced_at_least_once(historical)
            .await?;
 }
```

crates/fuel-streams-publisher/src/publisher/fuel_core_like.rs
Jurshsmith: Fwiw, I think we can remove the `await_synced_at_least_once` altogether. I did some local testing and it doesn't matter if the relayer has synced or not, `blocks_streams` sends the next block correctly. 

Also, if our last published == latest_block at point 0, that's fine, because we'll be using the BlockImporter's subscription then.

PR #379, opened 2025-01-16
feat(repo): Added load-tester and benchmarks

0xterminator:
Some changes to documentation too


benches/load-tester/src/runners/runner_streamable.rs
```diff
@@ -0,0 +1,35 @@
...
+    FuelStreams,
+};
+use fuel_streams_store::record::DataEncoder;
+use futures::StreamExt;
+
+use super::results::LoadTestTracker;
+
+pub async fn run_streamable_consumer<S: IntoSubject, T: DataEncoder>(
+    subject: S,
+    fuel_streams: Arc<FuelStreams>,
```
pedronauck, line 17: It would be nice to change here to use the `fuel-streams` crate directly instead of the NATS stream because if we test using the final Rust SDK, we will know more about how the entire stack is running instead of just NATS streams.

For that, you will need to add an environment variable to pass the WebServer URL so we can make these tests directly using the webserver deployed on staging.

pedronauck (changes requested):

PR #427, opened 2025-03-06
feat(repo): Added rest endpoints Postman documentation

0xterminator:


pedronauck (changes requested):
can you add a folder `postman/` at the root and move both collections for there? also rename here to have the be in `kebab-case` like the other?

PR #445, opened 2025-03-17
fix(sv-api): Rest API minor fixes

pedronauck:
## Tasks

- [x] Add `SwaggerUI` endpoint missed when migrated to Axum
- [x] Fix `utoipa` tag names to be according to Swagger
- [x] Remove `password_handler` and use just `api_key` handler when generating API keys
- [x] Move `/key/generate` endpoint from `sv-webserver` to `sv-api`


bacon.toml
```diff
@@ -0,0 +1,51 @@
+# This is a configuration file for the bacon tool
+#
+# Complete help on configuration: https://dystroy.org/bacon/config/
```
0xterminator, line 3: I dont know what bacon is but it looks good :)

0xterminator (approved):

rodrigobranas (approved):

PR #451, opened 2025-03-25
feat(repo): Add predicates 

pedronauck:


0xterminator (approved):
LGTM, nice code, most of the stuff we now have is already boilerplate. There is however some sections under assembly and loader_code that I am not very clear with yet, but overall it looks great.

PR #541, opened 2026-04-20
feat(docker): Add multi-platform support for Docker build

gondoi:


ironcev (dismissed):

gondoi:
Reviewed Bugbot's 4 findings on this PR — all anchored to commits (`09dfbca5`, `12c1ef02`, `f539c2e1`) that predate the final rewrite (`e4b3918`), which deletes the local composite action entirely and replaces the matrix-based job with one job per package using an explicit `if` (avoiding `matrix.*` in a reusable-workflow job-level `if`, which is exactly what the "Matrix context unavailable" finding warned about — already addressed, and the workflow even has a comment calling this out). imageid/QEMU/duplicate-platforms findings targeted the pre-rewrite composite action, which no longer exists. All 4 stale. No code changes needed.

gondoi:
Correction to my earlier comment: I lumped this finding in with the other 3 as "stale," but its \`original_commit_id\` is \`e4b3918\` — the actual HEAD, not a superseded intermediate commit. That was my mistake, not bugbot's.

On the substance: \`auth-mode: registry-login\` skips the OIDC/ECR-login steps in the shared composite entirely, and \`build-backend: native\` always hits the digest-push variant, which sets \`provenance: false\` — so nothing in the actual execution path calls for an OIDC token today. But I can't verify that by watching it run: this workflow has no \`pull_request\` trigger, so it's never exercised by this PR's CI. Every other repo on this migration using \`registry-login\` (prometheus-exporter, fuel-mcp-server, sway-playground, faucet) grants \`id-token: write\` anyway and all built green — this was the one outlier. Added it to match, rather than lean on an argument I can't test.

ironcev (approved):

